PR #1, opened 2019-06-19
ext: hal: add new API to get stm32 ipcc num of channel

arnop2:
Add API to read the The IPCC peripheral HWCFGR register and return the
number of channels capability.

Signed-off-by: Arnaud Pouliquen <arnaud.pouliquen@st.com>


erwango:
cc @nashif @galak

PR #12, opened 2019-06-24
stm32cube: L4: Re-enable CAN legacy driver

erwango:
Following recent L4 cube package update, activation of CAN
legacy driver was disabled by mistake and new CAN driver was
enabled.
Since zephyr STM32 CAN driver is still using legacy API,
re-enable legacy API.


erwango:
Tested in https://github.com/zephyrproject-rtos/zephyr/pull/16984

PR #19, opened 2019-07-30
HAL STM32: Synchronize STM Cube HSE clock value with zephyr clock value

Benichou34:
Synchronize STM Cube HSE clock value with zephyr clock value by defining HSE_VALUE as the same value of CONFIG_CLOCK_STM32_HSE_CLOCK.


erwango:
@galak , would you mind merging this change?

PR #22, opened 2019-10-06
stm32l151xb: add missing SPI_CR2_FRF definitions

karlp:
This is a bug in upstream CMSIS, but I don't have any contacts to get it
fixed there.

Signed-off-by: Karl Palsson <karlp@etactica.com>

### DIFF
--- a/stm32cube/stm32l1xx/README
+++ b/stm32cube/stm32l1xx/README
@@ -84,3 +84,10 @@ Patch List:
     Impacted files:
      drivers/include/stm32l1xx_hal_conf.h
     ST Bug tracker ID: NA. Not a stm32cube issue
+
+   *stm32l151xb: add missing SPI_CR2_FRF definitions
+
+    The higher density parts correctly included the SPI_CR2_FRF definitions,
+    but these were missing for the low/medium density parts.  Without these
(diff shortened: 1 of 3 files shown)


karlp:
I have attempted to do so for you, but given that I have no way of knowing those details ahead of time, I feel like this is something that project maintainers should do on merge.  Let me know if you want me to re-word it further.

erwango:
> I have attempted to do so for you, but given that I have no way of knowing those details ahead of time, I feel like this is something that project maintainers should do on merge. Let me know if you want me to re-word it further.

Sure, this was indeed waiting for my action.
Thanks for the update.

erwango:
@karlp, we had a late feedback from design to inform us that stm32l151xb stm32l151xa don't support this bit.
Can you confirm you actually need these definitions ?

karlp:
That's correct, but not actually relevant.  The user should not attempt to use this feature on a cat1/cat2 device, but the driver code attempts to _clear_ the bit, (safe and marked as zero on cat1/cat2) and without the definitions it simply doesn't even compile.  This was what my original commit message says.  unless you want to change the driver to only attempt to set/clear the bit based on the availability of the bit, but that's a different path altogether.

erwango:
Txs for clarification, I forgot about the context.

PR #29, opened 2019-12-04
stm32cube: all: Rename Kconfig symbol xxx_CRYPT to xxx_CRYP

erwango:
Each Cube peripheral API _foo.h has a matching Kconfig symbol
USE_STM32_HAL_FOO. This convention is useful for automatic package
update.
Crypto API derogates from this rule, which breaks automatic update:
- API is _cryp.h
- Kconfig is USE_STM32_HAL_CRYPT

Fix this by aligning Kconfig symbol on API naming.


galak:
Is there a matching change on the zephyr side?

erwango:
> Is there a matching change on the zephyr side?

Yeah it's coming. Had a coffee break in-bewteen :-)

PR #26, opened 2019-11-29
script: build new zephyr module from a new STM32Cube

FRASTM:
For stm32 series, get the version tree found on
https://github.com/STMicroelectronics/STM32Cube
and apply patches to update module to the latest version

run with `$ python ./scripts/update_stm32_package.py stm32xx`

Signed-off-by: Francois Ramu <francois.ramu@st.com>

PR #30, opened 2019-12-05
stm32cube: update packages to the minimal github ST repo versions

erwango:
STM32Cube packages are now available directly in github (cf https://github.com/STMicroelectronics/STM32Cube_MCU_Overall_Offer#stm32cube-mcu-packages)
From now on, these repo will be used for STM23Cube update. This is what is proposed by the update package script pushed in #26.

Since the script in #26 is doing a delta between previous version and new version, it expects that hal_stm32 packages version are available in github repos. It is already the case for most of the series but some series were not on par.
This commit series fixes that by updating hal_stm32 packages to the latest version available in github ST repos.

Commits in this PR will be included by a next PR which will do the update for other series using script provided in #26. And then we'll have only one PR in zephyr to do the full cube update.
Setting it as 'DNM' as it should not be merged by itself.


erwango:
@FRASTM

erwango:
Closing following merge of #33

PR #34, opened 2020-01-07
stm32cube: stm32g4xx: wrong clock setup

FRASTM:
When configuring the PLL > 80MHz, an intermediate step is needed 
with AHB prescaler set to 2 before setting the actual value.
In case the actual value is AHB prescaler 1, it must be set, though.

This will fix https://github.com/zephyrproject-rtos/zephyr/issues/21715

### DIFF
--- a/stm32cube/stm32g4xx/README
+++ b/stm32cube/stm32g4xx/README
@@ -35,4 +35,10 @@ License Link:
    https://opensource.org/licenses/BSD-3-Clause
 
 Patch List:
-   none
+
+   *stm32g4xx: wrong clock setup
+    In case the actual value is AHB prescaler 1, it must be set, 
+    after setting the AHB prescaler to 2.
+    Impacted files:
+     drivers/src/stm32g4xx_ll_utils.c
+    ST Bug tracker ID: 78880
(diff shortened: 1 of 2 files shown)


FRASTM:
stm32cube/stm32g4xx/README file updated

martinjaeger:
Just found that this PR fixed the LED blinky issue, but causes a wrong setting of `SysCoreClock` variable (75 MHz instead of actual 150 MHz frequency for the Nucleo STM32G431). This leads to wrong UART baud rate settings (that's how we discovered it) and possibly other issues...

In case of PLL frequency >80 MHz, the following call

https://github.com/zephyrproject-rtos/hal_stm32/blob/94d735c2613f5d17b64b78a15f0b0191e4474eb0/stm32cube/stm32g4xx/drivers/src/stm32g4xx_ll_utils.c#L325

sets also SysCoreClock to `hclk`, which is the preliminary lower AHB frequency, `pllfreq / 2`:
https://github.com/zephyrproject-rtos/hal_stm32/blob/94d735c2613f5d17b64b78a15f0b0191e4474eb0/stm32cube/stm32g4xx/drivers/src/stm32g4xx_ll_utils.c#L657

However, the call
https://github.com/zephyrproject-rtos/hal_stm32/blob/94d735c2613f5d17b64b78a15f0b0191e4474eb0/stm32cube/stm32g4xx/drivers/src/stm32g4xx_ll_utils.c#L331
afterwards only sets the AHB prescaler to final value, but does not update the `SysCoreClock` anymore.

@FRASTM can you please have a look? Thanks!

Can we try to get this fixed before Zephyr release 2.2 merge window closes? Otherwise the STM32G4 is hardly useable.

(posting here, as I can't open an issue in this repo).

PR #47, opened 2020-03-04
stm32cube: update stm32XX cube version

FRASTM:
stm32cube: update stm32f7 cube version
stm32cube: update stm32f4 cube version
stm32cube: update stm32g4 cube version
stm32cube: update stm32wb cube version
stm32cube: update stm32mp1 cube version
stm32cube: update stm32l4 cube version
stm32cube: update stm32h7 cube version


galak:
Should we split up the stm32 HAL module to match to git repos that now exist? @erwango thoughts?

erwango:
> Should we split up the stm32 HAL module to match to git repos that now exist? @erwango thoughts?

@galak We thought about it indeed. Obviously the upside is that we'd directly get the files we're interested in and no need to filter the package.
Issue I have in mind is that CMSIS and HAL packages have their own delivery rhythm and versioning. The Cube FW package ensures a coherency and compatibility between these packages. So we'd anyway need to access Cube FW delivery points to get the right versions of the packages that would ensure compatibility.

So for now, I prefer to stick with current solution.

FRASTM:
adding stm32f2  v1.9.0 
see https://github.com/zephyrproject-rtos/hal_stm32/pull/49

PR #50, opened 2020-04-15
stm32cube: rtc: Fix bit operations in LL_RTC_DATE_Config()

takumiando:
Currently, Zephyr doesn't need to apply this patch.
But it may need to do in the future.

Actually I need this patch for my own application :)


erwango:
Thanks @takumiando.
Maybe it would be better to push this change directly in STM32Cube repos:
https://github.com/STMicroelectronics/STM32CubeF4

takumiando:
@erwango Should I push to all STM32Cube series?

erwango:
> @erwango Should I push to all STM32Cube series?

This has to be discussed with maintainers. But at least push on the one series you're interested in.

takumiando:
Okay ;)

PR #53, opened 2020-05-10
stm32l0xx: Add stm32l0xx_ll_usb to CMakeLists

andysan:
Some STM32L0 SoCs contain a USB device controller. Add the LL USB HAL to enable USB support in Zephyr.


andysan:
This should be ready to be merged now.

PR #56, opened 2020-06-10
hal_stm32: add stm32h7xx_hal_rcc_ex.c

Nukersson:
Added missed stm32h7xx_hal_rcc_ex.c in CMakeLists.txt, which
may lead to undefined references
(e.g. HAL_RCCEx_GetD1SysClockFreq(...) )

### DIFF
--- a/stm32cube/stm32h7xx/CMakeLists.txt
+++ b/stm32cube/stm32h7xx/CMakeLists.txt
@@ -6,6 +6,7 @@
 zephyr_library_sources(soc/system_stm32h7xx.c)
 zephyr_library_sources(drivers/src/stm32h7xx_hal.c)
 zephyr_library_sources(drivers/src/stm32h7xx_hal_rcc.c)
+zephyr_library_sources(drivers/src/stm32h7xx_hal_rcc_ex.c)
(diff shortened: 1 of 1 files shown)


erwango:
> > This should not be needed.
> > Please enable CONFIG_USE_STM32_HAL_RCC_EX is you need this file in
> 
> Thank you. Probably providing some warning/service message with sufggestion to the user would be helpful.

@Nukersson Not sure how this can be done but we'll consider this, otherwise, please also consider this is documented in https://github.com/zephyrproject-rtos/hal_stm32/blob/master/README.rst.

Nukersson:
> > > This should not be needed.
> > > Please enable CONFIG_USE_STM32_HAL_RCC_EX is you need this file in
> > 
> > 
> > Thank you. Probably providing some warning/service message with sufggestion to the user would be helpful.
> 
> @Nukersson Not sure how this can be done but we'll consider this, otherwise, please also consider this is documented in https://github.com/zephyrproject-rtos/hal_stm32/blob/master/README.rst.

Thank you. I've found the line in CMakeLists.txt, where `CONFIG_USE_STM32_HAL_RCC_EX` is defined. I will close this PR. Additionally it can be redefined to `question`.

PR #65, opened 2020-09-05
Add support for automatic generation of die specific DT pinctrl files

gmarull:
This PR adds a new script to generate DT pinctrl files from CubeMX database. Generated files are compatible with DT bindings provided in https://github.com/zephyrproject-rtos/zephyr/pull/25996.

Highlights:
- Support for all STM32 series, including F1
- Configuration is stored in an external file
- Files are generated using Jinja2 templates
- Tests included
- Configuration for GitHub Actions (run tests and lint scripts) - sample run: https://github.com/gmarull/hal_stm32/actions/runs/244281534

Open questions:
- [x] Discuss if necessary adding defaults (e.g. `drive-push-pull`, `bias-disable`)
- [x] What do to with peripherals that may be configured depending on operation mode (e.g. SPI master/slave, low power modes...). Current approach is to use `variant` field in config file (see SPI example).
- [x] What to do with slew-rate (currently not supported). We need to be careful with defaults here, if layout was not intended for high speed usage it could cause EMI issues.
- [x] Grouping strategy, e.g. group UART/USART/LPUART or not (affects alphabetical order)
- [ ] Review drive/bias for all supported functions


scripts/genpinctrl/config.json
```diff
@@ -0,0 +1,219 @@
...
+        "stm32f3": "0x48000000",
+        "stm32f4": "0x40020000",
+        "stm32f7": "0x40020000",
+        "stm32g0": "0x50000000",
+        "stm32g4": "0x48000000",
+        "stm32h7": "0x58020000",
+        "stm32l0": "0x50000000",
+        "stm32l1": "0x40020000",
+        "stm32l4": "0x48000000",
+        "stm32l4p": "0x48000000",
```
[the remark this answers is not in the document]
gmarull: done

scripts/genpinctrl/config.json
```diff
@@ -0,0 +1,219 @@
...
+            "match": "^(?:LP)?US?ART\\d+_RTS$",
+            "drive": "open-drain",
+            "bias": "pull-up",
+            "mode": "output"
+        },
+        {
+            "name": "UART_TX / USART_TX / LPUART_TX",
+            "match": "^(?:LP)?US?ART\\d+_TX$",
+            "drive": "push-pull",
+            "bias": "pull-up",
```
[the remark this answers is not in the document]
gmarull: I think it's better to have a single configuration file, there'll be lots of duplicate content otherwise. Config file can be made more explicit for F1 only fields (e.g. `mode`) or non-F1 (e.g. `drive`). I still think it's better to keep things simple being a single family that diverges, it can be easily solved by an if/else in the Jinja template as it's already done for the include and pinmux entry.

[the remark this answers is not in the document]
gmarull: @erwango I have added conditional `bias` rendering depending on family and for F1, only render if `mode` is not `output` or `inout` (both map to `ALTERNATE` output). Can you check?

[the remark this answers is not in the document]
erwango: The way I see it is that user should be able, in this(these) config file(s), to see/set the description that he can see in the ref manual and expect in the dts files. 
The whole idea is to have something easy to deal with from the reference manual to the final pin configuration (and the other way round). The dts itself is just a intermediate step used as a normalized description.
I tried to follow this approach when working on the bindings, this is why we end up 2 different -pinctrl bindings (while I could have kept the same description and tweak the driver to abstract differences).

I would expect the pinctrl generation script allows to follow the same approach and keep configuration transparent to the user. So each one can verify that what is available in the config is aligned with  the ref man and the IP settings. This is why I'd prefer having 2 different config files when it makes sense, rather that tweaking the values in the script, even at cost of some config duplication.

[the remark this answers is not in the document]
gmarull: > The way I see it is that user should be able, in this(these) config file(s), to see/set the description that he can see in the ref manual and expect in the dts files.
> The whole idea is to have something easy to deal with from the reference manual to the final pin configuration (and the other way round). The dts itself is just a intermediate step used as a normalized description.
> I tried to follow this approach when working on the bindings, this is why we end up 2 different -pinctrl bindings (while I could have kept the same description and tweak the driver to abstract differences).
>
If we want to have comments in the config file we can switch to YAML, as JSON does not allow comments. I can take a look if you want, should be easy.
 
> I would expect the pinctrl generation script allows to follow the same approach and keep configuration transparent to the user. So each one can verify that what is available in the config is aligned with the ref man and the IP settings. This is why I'd prefer having 2 different config files when it makes sense, rather that tweaking the values in the script, even at cost of some config duplication.

Ok, I understand the idea. Let me see what I can do.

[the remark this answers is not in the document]
erwango: > > The way I see it is that user should be able, in this(these) config file(s), to see/set the description that he can see in the ref manual and expect in the dts files.
> > The whole idea is to have something easy to deal with from the reference manual to the final pin configuration (and the other way round). The dts itself is just a intermediate step used as a normalized description.
> > I tried to follow this approach when working on the bindings, this is why we end up 2 different -pinctrl bindings (while I could have kept the same description and tweak the driver to abstract differences).
> 
> If we want to have comments in the config file we can switch to YAML, as JSON does not allow comments. I can take a look if you want, should be easy. In fact, YAML may be more familiar to people because it's used in the DT bindings, we could also add the `description` field for detailed documentation.
 
The format itself sounds good to me (initially I don't have particular expectations on comments) but indeed yaml may be a good alternative.

> > I would expect the pinctrl generation script allows to follow the same approach and keep configuration transparent to the user. So each one can verify that what is available in the config is aligned with the ref man and the IP settings. This is why I'd prefer having 2 different config files when it makes sense, rather that tweaking the values in the script, even at cost of some config duplication.
> 
> Ok, I understand the idea. Let me see what I can do.

Great, txs

scripts/genpinctrl/config.json
```diff
@@ -0,0 +1,218 @@
...
+            "mode": "output",
+            "variant": "slave"
+        },
+        {
+            "name": "SPI_SLAVE_MOSI",
+            "match": "^SPI\\d+_MOSI$",
+            "drive": "push-pull",
+            "bias": "pull-down",
+            "mode": "input",
+            "variant": "slave"
```
erwango, line 139: I'm fine with the variant :-)

erwango:
Some ideas on the open points:

> - [ ] Discuss if necessary adding defaults (e.g. `drive-push-pull`, `bias-disable`)

This is not required. 
Tough, it could be nice, but in that case adding a comment such as `/* reset value */` would make it even nicer (if not too much trouble).

- [ ] What do to with peripherals that may be configured depending on operation mode (e.g. SPI master/slave, low power modes...). Current approach is to use `variant` field in config file (see SPI example).

I'm fine with the variant thing. Seems clear enough.

- [ ] What to do with slew-rate (currently not supported). We need to be careful with defaults here, if layout was not intended for high speed usage it could cause EMI issues.

I'd propose to adapt existing settings.
If needed, faster configs could be added using variants (for most common ones

- [ ] Grouping strategy, e.g. group UART/USART/LPUART or not (affects alphabetical order)

Current proposal seems fine.

- [ ] Review drive/bias for all supported functions

I'd propose to go with existing settings  as a first step. At least to verify that the new mechanism is not harmful.
In a second step, I'm fine with a global review.

gmarull:
> Some ideas on the open points:
> 
> > * [ ]  Discuss if necessary adding defaults (e.g. `drive-push-pull`, `bias-disable`)
> 
> This is not required.
> Tough, it could be nice, but in that case adding a comment such as `/* reset value */` would make it even nicer (if not too much trouble).
> 

@erwango should we then remove all `bias-disable` and `drive-push-pull` from config? I'll write some documentation to explain each entry, default if missing, etc.

scripts/genpinctrl/config.yaml
```diff
@@ -0,0 +1,158 @@
...
+      match: "^(?:LP)?US?ART\\d+_RX$"
+      mode: input
+
+    - name: USB_OTG_FS_DM
+      match: "^USB_OTG_FS_DM$"
+      mode: inout
+
+    - name: USB_OTG_FS_DP
+      match: "^USB_OTG_FS_DP$"
+      mode: inout
```
erwango, line 158: It is not clear what `inout` maps to within proposed dts bindings.

scripts/genpinctrl/config.yaml
```diff
@@ -0,0 +1,158 @@
...
+    stm32g4: 0x48000000
+    stm32h7: 0x58020000
+    stm32l0: 0x50000000
+    stm32l1: 0x40020000
+    stm32l4: 0x48000000
+    stm32l5: 0x42020000
+    stm32mp1: 0x50002000
+    stm32wb: 0x48000000
+
+alternate-functions:
```
erwango, line 19: `pin-config-templates` would be more in line with what we're intending to do here.

erwango, line 19: pin-config-templates would be more in line with what we're intending to do here.

scripts/genpinctrl/genpinctrl.py
```diff
@@ -0,0 +1,474 @@
...
+
+
+def format_mode(mode):
+    """Format mode for DT (F1 series).
+
+    Applied rules:
+        - Analog: Analog mode (ANALOG)
+        - Alternate function input: Input mode (GPIO_IN)
+        - Alternate function output: Alternate Function Output (ALTERNATE)
+        - Alternate function input/output: Alternate Function Output (ALTERNATE)
```
erwango, line 119: For ease of use, it would be good to be closer to F1 bindings here:
```
            * 0 : Alternate function output
            * 1 : General purpose output
            * 2 : Input
            * 3 : Analog
```

erwango (changes requested):

scripts/genpinctrl/config.yaml
gmarull: The purpose of the `mode` field is to specify if pin is operating as `analog`, `input`, `output` or `inout` (input/output, like for example I2C). This field is only relevant to F1 series, where the pin mode is configured using the following rules:

- Analog: Analog mode (ANALOG)
- Alternate function input: Input mode (GPIO_IN)
- Alternate function output: Alternate Function Output (ALTERNATE)
- Alternate function input/output: Alternate Function Output (ALTERNATE)

It's basically needed for cases like `UART_RX`. Even though an alternate function is used, pin needs to be configured as `GPIO_IN`. This is what I understood from F1 datasheet.

scripts/genpinctrl/genpinctrl.py
gmarull: This function maps F1 mode depending on operation mode, which follows the rules mentioned in the comment of the function.

scripts/genpinctrl/config.yaml
erwango: Ok if this field is f1 only. So maybe for clarity it could be renamed as `f1-mode` or alike.

Though, on F1 there is no input/output mode. From §9.1 'GPIO functional description'
Screenshot from 2020-09-21 10-02-47
So, looking to proposed mapping from an external point of  view:
`analog` mode also translates to `ANALOG` (ok)
`input` mode translates to `GPIO_IN` (ok)
`output` mode translates to `ALTERNATE` (why not using `alternate` then.. but ok)
`inout` mode also translates to `ALTERNATE` (wait!?)
And we have no way to select `General Purpose output`, which I agree is not stricly needed, but it is available in dts binding and then we can get easily confused with the `ouput` one.

So at the end, I really consider that the mixing of F1 and other series makes things unclear.

Taking the example of UART_TX, the config.yaml proposes:
```
    - name: UART_TX / USART_TX / LPUART_TX
      match: "^(?:LP)?US?ART\\d+_TX$"
      bias: pull-up
      mode: output
```
For STM32F1, Reference manual mentions it should be "Alternate function push-pull".
It also adds: "The weak pull-up and pull-down resistors are disabled.", but looking at the config:
-`      mode: output`
 F1 only, but `output` translate to `ALTERNATE` in the end.
-`      bias: pull-up`
`pull-up` applies for all series, but for that case on F1, it doesn't apply (why it would have applied on an input mode for instance).

So at the end, since we're mixing series,
- on F1, it would happen that `bias` is specified, but not applied 
- for the F1 only `mode`, it translates to a different value (`output` > `alternate`)

I agree that, at the end, the correct configuration is generated.
But for someone looking the config from f1 point of view, it could seems quite unclear, or even buggy for someone that haven't read the reference manual.

If we're adding to that the fact that the default configuration are not mentioned (which is a choice that I totally support), this increases confusion to a higher level.

So in the end I really like the direction this is going to but this F1/other mixing is the dark point that we'd need to sort out. I agree this will have a cost of config file duplication, but will be much easier to understand for day to day and one time users.

scripts/genpinctrl/config-f1.yaml
```diff
@@ -0,0 +1,156 @@
...
+- name: SPI_SLAVE_MOSI
+  match: "^SPI\\d+_MOSI$"
+  bias: pull-down
+  mode: input
+  variant: slave
+
+- name: SPI_SLAVE_SCK
+  match: "^SPI\\d+_SCK$"
+  bias: pull-down
+  mode: input
```
erwango, line 109: Ref man recommends: "Input floating"
Current zephyr config is (STM32_MODE_INPUT | STM32_CNF_IN_FLOAT)

scripts/genpinctrl/config-f1.yaml
```diff
@@ -0,0 +1,156 @@
...
+
+- name: SPI_SLAVE_MISO
+  match: "^SPI\\d+_MISO$"
+  mode: alternate
+  variant: slave
+
+- name: SPI_SLAVE_MOSI
+  match: "^SPI\\d+_MOSI$"
+  bias: pull-down
+  mode: input
```
erwango, line 103: Refman recommends "Input floating / Input pull-up"
Current code: `#define STM32_PIN_SPI_SLAVE_MOSI	(STM32_MODE_INPUT | STM32_CNF_IN_FLOAT)`

scripts/genpinctrl/config-f1.yaml
```diff
@@ -0,0 +1,156 @@
...
+  mode: alternate
+
+- name: TIM_BKIN
+  match: "^TIM\\d+_BKIN\\d?$"
+  bias: pull-up
+  mode: input
+
+- name: TIM_ETR
+  match: "^TIM\\d+_ETR$"
+  mode: input
```
erwango, line 129: I propose that for now we restrict pin timer configuration to pwm for now (using variant)

scripts/genpinctrl/config-f1.yaml
```diff
@@ -0,0 +1,156 @@
...
+- name: UART_CTS / USART_CTS
+  match: "^US?ART\\d+_CTS$"
+  drive: open-drain
+  bias: pull-up
+  mode: input
+
+- name: UART_RTS / USART_RTS
+  match: "^US?ART\\d+_RTS$"
+  drive: open-drain
+  mode: alternate
```
erwango, line 140: According to refman: "Alternate function push-pull"

scripts/genpinctrl/config-f1.yaml
```diff
@@ -0,0 +1,156 @@
...
+  match: "^US?ART\\d+_RX$"
+  mode: input
+
+- name: USB_OTG_FS_DM
+  match: "^USB_OTG_FS_DM$"
+  mode: alternate
+
+- name: USB_OTG_FS_DP
+  match: "^USB_OTG_FS_DP$"
+  mode: alternate
```
erwango, line 156: Here refman says: "Controlled automatically by the USB power-down".
Current code: `#define STM32_PIN_USB			(STM32_MODE_INPUT | STM32_CNF_IN_PUPD)`

Not sure what to do here.

scripts/genpinctrl/config-f1.yaml
```diff
@@ -0,0 +1,156 @@
...
+#
+# This file contains a list of pin configuration templates used to generate the
+# pinctrl files. Each entry can have the following fields:
+#
+#   - name (mandatory): This is the pin alternate function name, e.g. UART_TX.
+#     It is used to group pin configurations alphabetically in the generated
+#     pinctrl files.
+#
+#   - match (mandatory): This is a regular expression used to match against
+#     CubeMX pin configuration names. The regular expression should be as
```
erwango, line 11:
```suggestion
#     STM32 xml database pin configuration names. The regular expression should be as
```

scripts/genpinctrl/config.yaml
```diff
@@ -0,0 +1,139 @@
...
+#
+# This file contains a list of pin configuration templates used to generate the
+# pinctrl files. Each entry can have the following fields:
+#
+#   - name (mandatory): This is the pin alternate function name, e.g. UART_TX.
+#     It is used to group pin configurations alphabetically in the generated
+#     pinctrl files.
+#
+#   - match (mandatory): This is a regular expression used to match against
+#     CubeMX pin configuration names. The regular expression should be as
```
erwango, line 11:
```suggestion
#     STM32 xml database pin configuration names. The regular expression should be as
```

scripts/genpinctrl/config.yaml
```diff
@@ -0,0 +1,139 @@
...
+#     precise as possible. Note that it needs to be escaped here in the
+#     configuration file.
+#
+#   - bias (optional): Bias setting (disable, pull-up, pull-down). If not set
+#     pinctrl file will not contain the bias entry, equivalent to disable
+#     setting.
```
erwango, line 17:
```suggestion
#   - bias (optional): Bias setting (disable, pull-up, pull-down). This setting
#     only applies to "input" mode. Equivalent to "disable" (a.k.a floating) if not. 
```

erwango (changes requested):
First batch of comments on F1 (pushing before loosing them). Nothing critical.

scripts/genpinctrl/config.yaml
```diff
@@ -0,0 +1,139 @@
...
+#
+#   - drive (optional): Drive setting (push-pull, open-drain). If not set
+#     pinctrl file will not contain the drive entry, equivalent to push-pull
+#     setting
```
erwango, line 21:
```suggestion
#   - drive (optional): Drive setting (push-pull, open-drain). Equivalent to "push-pull" if not set.
```

scripts/genpinctrl/config.yaml
```diff
@@ -0,0 +1,139 @@
...
+#
+#   - variant (optional): Define another variant of a pin configuration. This
+#     setting can be useful to define multiple configurations for a certain
+#     alternate function.
```
erwango, line 25:
```suggestion
#   - variant (optional): Defines an alternative pin configuration. This is used
#     to provide multiple configurations of a pin function (slave, master, low-power, ...)
```

scripts/genpinctrl/config.yaml
```diff
@@ -0,0 +1,139 @@
...
+- name: CAN_RX
+  match: "^CAN\\d+_RX$"
+  bias: pull-up
+
+- name: CAN_TX
+  match: "^CAN\\d+_TX$"
+
+- name: DAC_OUT
+  match: "^DAC(?:\\d+)?_OUT\\d+$"
+
```
erwango, line 41: I guess the analog is selected through the script.
Would there be a way to add a mode similarly to f1 ?
Reason is that CAN_TX has no option, meaning Alternate/Push-Pull/No Pull
Here no option leads to analog...

scripts/genpinctrl/config.yaml
```diff
@@ -0,0 +1,139 @@
...
+- name: FDCAN_RX
+  match: "^FDCAN\\d+_RX$"
+  bias: pull-up
```
erwango, line 44: Current code is `STM32_PUSHPULL_NOPULL`

scripts/genpinctrl/config.yaml
```diff
@@ -0,0 +1,139 @@
...
+  variant: master
+
+- name: SPI_MASTER_MOSI
+  match: "^SPI\\d+_MOSI$"
+  bias: pull-down
+  variant: master
+
+- name: SPI_MASTER_SCK
+  match: "^SPI\\d+_SCK$"
+  bias: pull-down
```
erwango, line 80: Current code: `STM32_PUSHPULL_NOPULL`

scripts/genpinctrl/config.yaml
```diff
@@ -0,0 +1,139 @@
...
+
+- name: SPI_SLAVE_SCK
+  match: "^SPI\\d+_SCK$"
+  bias: pull-down
+  variant: slave
+
+- name: SPI_SLAVE_NSS
+  match: "^SPI\\d+_NSS$"
+  bias: pull-up
+  variant: slave
```
erwango, line 106: Is it worth adding slave/master config since there is no diff?

scripts/genpinctrl/config.yaml
```diff
@@ -0,0 +1,139 @@
...
+
+- name: TIM_CH / TIM_CHN
+  match: "^TIM\\d+_CH\\d+N?$"
+
+- name: TIM_BKIN
+  match: "^TIM\\d+_BKIN\\d?$"
+  bias: pull-up
+
+- name: TIM_ETR
+  match: "^TIM\\d+_ETR$"
```
erwango, line 116: I propose to add these later

scripts/genpinctrl/genpinctrl.py
```diff
@@ -0,0 +1,494 @@
+"""
+Utility to autogenerate Zephyr DT pinctrl files for all STM32 microcontrollers.
+
+Usage::
+
+    python genpinctrl.py -p /path/to/STM32CubeMX -o /path/to/output_dir
```
erwango, line 6:
```suggestion
    python3 genpinctrl.py -p /path/to/STM32CubeMX -o /path/to/output_dir
```

erwango (changes requested):

scripts/genpinctrl/config.yaml
gmarull: F1 series need different configs for master/slave, as MISO/MOSI will be either alternate/input depending on operation mode. Other series don't need that, what should we do?

erwango: This is one of the interest of separating the config files. I think they should do whatever makes sense:
If no variant is needed, let's not add it.

scripts/genpinctrl/config-f1.yaml
```diff
@@ -0,0 +1,156 @@
...
+#     precise as possible. Note that it needs to be escaped here in the
+#     configuration file.
+#
+#   - mode (mandatory): Mode setting (analog, alternate, input). Mode needs to
+#     be set according to the following rules:
+#       * Pin operates in analog mode: analog
+#       * Pin operates as an alternate input: input
+#       * Pin operates as an alternate output: alternate
+#       * Pin operates as an alternate bidirectional: alternate
```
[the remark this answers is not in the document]
gmarull: fixed

scripts/genpinctrl/config-f1.yaml
```diff
@@ -0,0 +1,156 @@
...
+#
+#   - bias (optional): Bias setting (disable, pull-up, pull-down). This setting
+#     only applies to input mode. If not set pinctrl file will not contain the
+#     bias entry, equivalent to disable setting.
+#
+#   - drive (optional): Drive setting (push-pull, open-drain). If not set
+#     pinctrl file will not contain the drive entry, equivalent to push-pull
+#     setting
```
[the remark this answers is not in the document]
gmarull: fixed

scripts/genpinctrl/config-f1.yaml
```diff
@@ -0,0 +1,156 @@
+# Configuration file for pinctrl generation (F1 series only).
+#
+# This file contains a list of pin configuration templates used to generate the
+# pinctrl files. Each entry can have the following fields:
+#
+#   - name (mandatory): This is the pin alternate function name, e.g. UART_TX.
```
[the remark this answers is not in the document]
gmarull: fixed

scripts/genpinctrl/config-f1.yaml
```diff
@@ -0,0 +1,156 @@
...
+#     only applies to input mode. If not set pinctrl file will not contain the
+#     bias entry, equivalent to disable setting.
+#
+#   - drive (optional): Drive setting (push-pull, open-drain). If not set
+#     pinctrl file will not contain the drive entry, equivalent to push-pull
+#     setting
+#
+#   - variant (optional): Define another variant of a pin configuration. This
+#     setting can be useful to define multiple configurations for a certain
+#     alternate function.
```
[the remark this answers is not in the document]
gmarull: fixed

scripts/genpinctrl/config-f1.yaml
gmarull: fixed

scripts/genpinctrl/config.yaml
gmarull: fixed

scripts/genpinctrl/config.yaml
gmarull: There is no "input" mode for non-F1 series, so I removed `This setting only applies to "input" mode`

scripts/genpinctrl/config.yaml
gmarull: fixed

scripts/genpinctrl/config.yaml
gmarull: In case of non-F1 series a pin will be configured either as alternate or analog. Using XML database this setting can be fully determined, so a `mode` field will be redundant. F1 is special because it requires the "input" case.

scripts/genpinctrl/config.yaml
gmarull: Should `CAN_RX` bias be removed then?

scripts/genpinctrl/config.yaml
gmarull: fixed

scripts/genpinctrl/config.yaml
gmarull: I have added `SPI_MISO/MOSI/SCK/NSS` for non-F1 and master/slave variants for F1, let me know if it's ok now.

scripts/genpinctrl/genpinctrl.py
gmarull: fixed

scripts/genpinctrl/config.yaml
gmarull: fixed

scripts/genpinctrl/config-f1.yaml
```diff
@@ -0,0 +1,156 @@
...
+  mode: alternate
+
+- name: I2S_SD
+  match: "^I2S\\d+_SD$"
+  mode: alternate
+
+- name: SPI_MASTER_MISO
+  match: "^SPI\\d+_MISO$"
+  bias: pull-down
+  mode: input
```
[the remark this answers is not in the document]
gmarull: fixed

scripts/genpinctrl/config-f1.yaml
```diff
@@ -0,0 +1,156 @@
...
+#   - mode (mandatory): Mode setting (analog, alternate, input). Mode needs to
+#     be set according to the following rules:
+#       * Pin operates in analog mode: analog
+#       * Pin operates as an alternate input: input
+#       * Pin operates as an alternate output: alternate
+#       * Pin operates as an alternate bidirectional: alternate
+#
+#   - bias (optional): Bias setting (disable, pull-up, pull-down). This setting
+#     only applies to input mode. If not set pinctrl file will not contain the
+#     bias entry, equivalent to disable setting.
```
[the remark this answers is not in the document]
gmarull: fixed

scripts/genpinctrl/config-f1.yaml
gmarull: fixed

scripts/genpinctrl/config.yaml
gmarull: There's an interesting point: analog mode seems to be used for low-power settings, see e.g. https://wiki.st.com/stm32mpu/wiki/LPTIM_device_tree_configuration. If this is true, then a `mode` field is required.

erwango: > Using XML database this setting can be fully determined, so a mode field will be redundant.

Sure, but this is confusing for the reader since two similar pin configurations lead to different pinctrls in the end.
Maybe a comment field (discarded at processing) would be useful for clarity.

> If this is true, then a mode field is required.

Right, we'll need this field for lp variants when we need to force pins to analog, and potentially other usages.

erwango (changes requested):

scripts/genpinctrl/config.yaml
erwango: This is fine, txs

scripts/genpinctrl/config.yaml
erwango: for CAN_RX, we have both PULLUP and NOPULL in the code right now.
Let's see what users report during review.

scripts/genpinctrl/config.yaml
gmarull: Added `mode` field

scripts/genpinctrl/config.yaml
```diff
@@ -0,0 +1,136 @@
...
+  bias: pull-down
+
+- name: SPI_MOSI
+  match: "^SPI\\d+_MOSI$"
+  mode: alternate
+  bias: pull-down
+
+- name: SPI_SCK
+  match: "^SPI\\d+_SCK$"
+  mode: alternate
```
erwango, line 95: To match with current code configuration:
```suggestion
- name: SPI_SCK
  match: "^SPI\\d+_SCK$"
  mode: alternate
  slew-rate: "very-high-speed"
```

erwango (changes requested):

erwango:
^^ @ABOSTM, @FRASTM, @arnopo, @Mani-Sadhasivam, @ydamigos, @cybertale, @lowlander, @Nukersson, @lochej, @avisconti, @loicpoulain
Please have a look on this proposition. Aim of this change (to be used along with https://github.com/zephyrproject-rtos/zephyr/pull/25996) is to replace and finally remove the pinmux.c files and from now on configure device peripheral pins using only device tree.

scripts/genpinctrl/config.yaml
gmarull: Do MISO/MOSI require `very-high-speed`?

erwango: No, only SCK for most cases.

gmarull: As a bit per clock cycle is transmitted, slew-rate should ideally be the same on all SPI lines. For very high clock rates we could potentially see degraded data signals w.r.t SCK. CubeMX generates the following when selecting SPI:
```c
    /**SPI1 GPIO Configuration    
    PB5     ------> SPI1_MOSI
    PB4     ------> SPI1_MISO
    PB3     ------> SPI1_SCK
    PA15     ------> SPI1_NSS 
    */
    GPIO_InitStruct.Pin = GPIO_PIN_5|GPIO_PIN_4|GPIO_PIN_3;
    GPIO_InitStruct.Mode = GPIO_MODE_AF_PP;
    GPIO_InitStruct.Pull = GPIO_NOPULL;
    GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_VERY_HIGH;
    GPIO_InitStruct.Alternate = GPIO_AF5_SPI1;
    HAL_GPIO_Init(GPIOB, &GPIO_InitStruct);

    GPIO_InitStruct.Pin = GPIO_PIN_15;
    GPIO_InitStruct.Mode = GPIO_MODE_AF_PP;
    GPIO_InitStruct.Pull = GPIO_NOPULL;
    GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_VERY_HIGH;
    GPIO_InitStruct.Alternate = GPIO_AF5_SPI1;
    HAL_GPIO_Init(GPIOA, &GPIO_InitStruct);
```

erwango: Current state of zephyr code is the result of what has been used, tested, reported failed and fixed on Zephyr enabled STM32 targets. So for now, I prefer we stick with same configuration used in Zephyr.

erwango: But I'm open to change that in a latter step.

galak (commented):
Its a nit, but is it possible to remove the blank line after the comment:

```
			/* I2C_SDA */

			i2c1_sda_pa10: i2c1_sda_pa10 {
```

should be:

```
			/* I2C_SDA */
			i2c1_sda_pa10: i2c1_sda_pa10 {
```

scripts/genpinctrl/config.yaml
gmarull: As noted in current pinmux files (F4), this seems to be a requirement to overcome errata:

image

I wonder if there should be a high speed `variant` for SPI_SCK for these cases. Using high slew-rates may lead to EMI problems, reflections, etc. on boards that are not properly routed (because they may not need such high speeds).

ABOSTM (approved):
LGTM

scripts/genpinctrl/config.yaml
erwango: Not sure about the speed variant. I agree this would make sense in this specific case, but I'd like to avoid having people adding their own variant for any reason.
Over variant, I'd prefer to overwrite the pin node directly in board dts in a first time. So it is done exactly in the few cases it is required.
In both cases, drawback is that people have to know he issue first before fixing it (rather than using directly the default configuration that fixes the issue). But we can have a try and see how people react (we have a full length release to get this fixed).

erwango (approved):

Nukersson:
Is it possible just to run the script, which creates corresponding *.dtsi files at Zephyr build time? 

Other suggestions would be:
 - Providing *.dtsi generator script with Zephyr and adding notice to Zephyr's documentation for STM32CubeMX installation and generator script usage.

Nukersson (changes requested):
My opinion here is: `flake8` and `black` tools are belonging to separate one PR.

erwango:
> Is it possible just run the *.dtsi gen script at `Zephyr build` or `west update`, or by user's action once for each particular SoC?
> The advantage here would be that we need just test the script itself, but not each particular dtsi file.

If we don't provide the *.dtsi, we still need to provide the script input (*.xml files from CubeMx database). Since it's not possible to have zephyr build depending on outside component (CubeMx installation in that case), we'd need a zephyr module containing CubeMx database. So instead of having *.dtsi files, you'd have to pull each *.xml (which is the same size (or larger) but harder to read). I'm not sure there's any gain here.
Also, generation would have to be done for each build, which might be harmless for a single build, but more impacting in CI (eg: 14298 builds for last night full STM32 daily non reg).

Last, I personally find interesting having the complete *.pinctrl.dtsi file for each SoC, as, assuming it is correct (which should be the case given the source is correct), this is an easy documentation to parse and read to find an appropriate pin mapping.

gmarull:
> My opinion here is: `flake8` and `black` tools are belonging to separate one PR.

These are tools optionally used for development and have been added in a specific commit. `flake8` is also run by the GH Actions script in order to make sure scripts are linted before being merged. It would probably be good to document that part.

Nukersson:
@gmarull Thank you for this PR and answering my questions (you 2 @erwango). From my side I can test only H745ZI SoC, since I have currently only this one. I will drop my test results here later.

PR #66, opened 2020-10-01
scripts: genpinctrl: handle pin names with extra description

gmarull:
Some pin names have another name appended, e.g. "PA15 (JTDI)".  Current regex did not handle those cases, so it has been adjusted to handle them.

### DIFF
--- a/dts/st/f0/stm32f030cctx-pinctrl.dtsi
+++ b/dts/st/f0/stm32f030cctx-pinctrl.dtsi
@@ -72,6 +72,12 @@
 				drive-open-drain;
 			};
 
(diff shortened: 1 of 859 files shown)


lochej:
> Some pin names have another name appended, e.g. "PA15 (JTDI)". Current regex did not handle those cases, so it has been adjusted to handle them.

Thanks for this fix ! I have also just noticed the reason and was gonna inform you about it :) Nice we could get a fix quickly

galak (approved):

lochej (approved):
LGTM Thanks !

erwango (approved):

PR #67, opened 2020-10-02
Document new dts directory and related files

erwango:
Update repo doc to feature new generated dtsi files.

Also contain minor changes to generation scripts.


erwango:
^^gmarull

scripts/genpinctrl/genpinctrl.py
```diff
@@ -444,8 +444,10 @@ def main(cube_path, output):
     mcu_signals = get_mcu_signals(cube_path, gpio_ip_afs)
 
     if os.path.exists(output):
-        shutil.rmtree(output)
-    os.makedirs(output)
+        for dirpath, dirnames, filenames in os.walk(output):
+            # Remove directories, ignore files (README)
+            for dirname in dirnames:
+                shutil.rmtree(os.path.join(dirpath, dirname))
```
gmarull, line 450: there should be an else case here creating the output directory in case it doesn't exist. Even though it will never happen now in the repo, it may be needed for other paths.

erwango: Indeed, I haven't though of other paths, I'll fix that.

erwango: Fixed

gmarull (approved):

galak (approved):

erwango:
@galak Do we need a matching a PR for this in Zephyr or can we merge it and it will be integrated in the flow.

galak:
> @galak Do we need a matching a PR for this in Zephyr or can we merge it and it will be integrated in the flow.

I think its ok to merge this w/o an upstream PR.

erwango:
^@ABOSTM

ABOSTM (commented):
LGTM

PR #69, opened 2020-10-08
pins: dts: Add missing USB_OTG_FS_ID pin

galak:
Add USB_OTG_FS_ID to config.yaml and re-gen dts pinctrl.dtsi files

Signed-off-by: Kumar Gala <kumar.gala@linaro.org>

### DIFF
--- a/dts/st/f2/stm32f205r(b-c-e-f-g)tx-pinctrl.dtsi
+++ b/dts/st/f2/stm32f205r(b-c-e-f-g)tx-pinctrl.dtsi
@@ -797,6 +797,13 @@
 				pinmux = <STM32_PINMUX('A', 12, AF10)>;
 			};
 
+			/* USB_OTG_FS_ID */
+
+			usb_otg_fs_id_pa10: usb_otg_fs_id_pa10 {
+				pinmux = <STM32_PINMUX('A', 10, AF10)>;
+				bias-pull-up;
+			};
+
 		};
 	};
 };
--- a/dts/st/f2/stm32f205r(e-g)yx-pinctrl.dtsi
+++ b/dts/st/f2/stm32f205r(e-g)yx-pinctrl.dtsi
@@ -797,6 +797,13 @@
 				pinmux = <STM32_PINMUX('A', 12, AF10)>;
 			};
 
+			/* USB_OTG_FS_ID */
+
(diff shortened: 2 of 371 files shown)


gmarull (changes requested):
- F1 config file should also be updated
- It looks like `pull-up` is necessary: https://github.com/zephyrproject-rtos/zephyr/blob/master/drivers/pinmux/stm32/pinmux_stm32l5x.h#L37

galak:
> * F1 config file should also be updated
> * It looks like `pull-up` is necessary: https://github.com/zephyrproject-rtos/zephyr/blob/master/drivers/pinmux/stm32/pinmux_stm32l5x.h#L37

for F1 is I set `pull-up` I get:

`ValueError: Bias can only be set for input mode on F1 (entry: USB_OTG_FS_ID)`

gmarull:
> for F1 is I set `pull-up` I get:
> 
> `ValueError: Bias can only be set for input mode on F1 (entry: USB_OTG_FS_ID)`

F1 has some special rules compared to the rest of families (see the docs in the F1 config file). In this case it should be configured as an `input` + `pull-up`.

galak:
> F1 has some special rules compared to the rest of families (see the docs in the F1 config file). In this case it should be configured as an `input` + `pull-up`

Fixed.

gmarull:
@galak seems like script needs to be re-run, no changes are present for F1

galak:
> @galak seems like script needs to be re-run, no changes are present for F1

I re-ran, doesn't change anything for F1.

gmarull:
@galak all USB stuff from F1 config file can actually be deleted, it's not used.

galak:
> @galak all USB stuff from F1 config file can actually be deleted, it's not used.

Updated to remove USB stuff from f1.

gmarull (approved):

erwango (changes requested):
Please split the initial commit in 2, so that scripts/genpinctrl/config.yaml could be reviewed easily.

@gmarull Maybe we'll need to add a GH Action to check this ?

galak:
> Please split the initial commit in 2, so that scripts/genpinctrl/config.yaml could be reviewed easily.
> 
> @gmarull Maybe we'll need to add a GH Action to check this ?

Sure thing.

erwango (approved):

PR #70, opened 2020-10-08
pins: dts: Fix regex to support CAN_{TX,RX} and CANn_{TX,RX}

galak:
Some SoCs have the CAN signals without a number identifier (CAN_TX,
CAN_RX) so tweak regex to support with and without a number id.

Signed-off-by: Kumar Gala <kumar.gala@linaro.org>

### DIFF
--- a/dts/st/f0/stm32f042c(4-6)tx-pinctrl.dtsi
+++ b/dts/st/f0/stm32f042c(4-6)tx-pinctrl.dtsi
@@ -52,6 +52,28 @@
 				pinmux = <STM32_PINMUX('B', 1, ANALOG)>;
 			};
 
+			/* CAN_RX */
+
+			can_rx_pa11: can_rx_pa11 {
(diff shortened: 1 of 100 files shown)


gmarull (approved):

erwango (approved):
I'm fine with the change.
One thing is that introduces a divergence between nodes labels (`can1`) and pinctrl ones (`can_foo`), but I don't think this is too much trouble.
I let @martinjaeger comment on that regard.

martinjaeger:
> One thing is that introduces a divergence between nodes labels (`can1`) and pinctrl ones (`can_foo`), but I don't think this is too much trouble.

I think that's fine. It's consistent with other peripherals for STM32, e.g. the DAC. If the MCU series has only one DAC, we also use `dac_` in the pinctrl and `dac1` as the devicetree node.

martinjaeger (approved):

PR #73, opened 2020-10-12
scripts: genpinctrl: fix multiple F1 issues

gmarull:
When working with F1 series pins not present in the GPIO IP files need to be taken into account with the `NO_REMAP` setting. Furthermore, only first remap block was being processed, now all of them are taken into account.

Fixes https://github.com/zephyrproject-rtos/zephyr/issues/29120


erwango (approved):

erwango (changes requested):
s/FULL_REMAP/REMAP_FULL  (according to `include/dt-bindings/pinctrl/stm32f1-pinctrl.h`)

gmarull:
> s/FULL_REMAP/REMAP_FULL (according to `include/dt-bindings/pinctrl/stm32f1-pinctrl.h`)

Good catch, fixed

erwango (approved):

PR #74, opened 2020-10-15
Add support for Ethernet pins

gmarull:
F1 settings taken from RM0008 Rev 20, Table 209.
Other series settings are taken from current Zephyr settings. Note that
they will likely need to be adjusted, as very-high-speed is not needed
for all signals.

### DIFF
--- a/scripts/genpinctrl/stm32-pinctrl-config.yaml
+++ b/scripts/genpinctrl/stm32-pinctrl-config.yaml
@@ -48,6 +48,106 @@
   match: "^DAC(?:\\d+)?_OUT\\d+$"
   mode: analog
 
+- name: ETH_COL
+  match: "^ETH_COL$"
+  mode: alternate
+  slew-rate: very-high-speed
+
+- name: ETH_CRS
+  match: "^ETH_CRS$"
+  mode: alternate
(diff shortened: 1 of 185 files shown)


scripts/genpinctrl/stm32-pinctrl-config.yaml
```diff
@@ -48,6 +48,95 @@
...
+- name: ETH_TXD2
+  match: "^ETH_TXD2$"
+  mode: alternate
+  slew-rate: high-speed
+
+- name: ETH_TXD3
+  match: "^ETH_TXD3$"
+  mode: alternate
+  slew-rate: high-speed
+
```
erwango, line 139: In current version of the pinmux setting, all of these are `STM32_OSPEEDR_VERY_HIGH_SPEED`.
I'm not opposed ti fine tune this, but I'd prefer that we do first the transition "iso-config", so it can be seen that dts pinctrl has no functional impact.
And in a second step, we'll be able to come back on these and fin tune.

erwango (changes requested):

erwango (approved):

galak (approved):

PR #77, opened 2020-10-22
dts: Provide signals for SDMMC

erwango:
Configuration and updated -pinctrl.dtsi files for SDMMC signals

### DIFF
--- a/scripts/genpinctrl/stm32-pinctrl-config.yaml
+++ b/scripts/genpinctrl/stm32-pinctrl-config.yaml
@@ -180,6 +180,11 @@
   match: "^I2S\\d+_SD$"
   mode: alternate
 
+- name: SDMMC
+  match: "^SDMMC\\d+_(?:CK)?(?:CKIN)?(?:CDIR)?(?:CMD)?(?:D\\d+)?(?:D0DIR)?(?:D123DIR)?$"
+  mode: alternate
+  slew-rate: very-high-speed
+
 - name: SPI_MISO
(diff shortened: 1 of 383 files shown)


scripts/genpinctrl/stm32-pinctrl-config.yaml
```diff
@@ -180,6 +180,11 @@
   match: "^I2S\\d+_SD$"
   mode: alternate
 
+- name: SDMMC
+  match: "^SDMMC\\d+_(?:CK)?(?:CKIN)?(?:CDIR)?(?:CMD)?(?:D\\d+)?(?:D0DIR)?(?:D123DIR)?$"
+  mode: alternate
+  slew-rate: very-high-speed
```
gmarull, line 186: It may be better to split them in case slew-rate doesn't need to be `very-high-speed` for all the signals.

erwango: I agree this could be done later on if required, but I opted for simplicity for now.

gmarull: Note that if they need to be split, the `name` will also need to change, so the diff will be larger than if they are split from the beginning.

erwango: Yeah I thought about that, but even if they had to be split, likely they'll be split  in groups.. and for now I don't know which groups..

gmarull (approved):

galak (approved):

PR #79, opened 2020-10-26
scripts/genpinctrl:  Add generation of "additional functions" for non F1 series

erwango:
Fixes https://github.com/zephyrproject-rtos/zephyr/issues/29368

Generate signals for "additional functions" for pins on non F1 series.
Set mode as an optional parameters and by default infer its value (analog/alternate) from the result of the input files parsing.
For  a given pin signal, if an alternate function is found, consider mode as "alternate".
If not AF is found,consider the signal as valid, with mode as "analog". This matches the definition of "additional functions"
that could be found in ref manual and datasheets.

### DIFF
--- a/scripts/genpinctrl/genpinctrl.py
+++ b/scripts/genpinctrl/genpinctrl.py
@@ -80,15 +80,15 @@ def validate_config_entry(entry, family):
     if not entry.get("match"):
         raise ValueError(f"Missing entry match for {entry['name']}")
 
-    if not entry.get("mode"):
-        raise ValueError(f"Missing entry mode for {entry['name']}")
-
     if family == "STM32F1":
+        if not entry.get("mode"):
+            raise ValueError(f"Missing entry mode for {entry['name']}")
         if entry["mode"] not in ("analog", "input", "alternate"):
             raise ValueError(f"Invalid mode for {entry['name']}: {entry['mode']}")
     else:
-        if entry["mode"] not in ("analog", "alternate"):
-            raise ValueError(f"Invalid mode for {entry['name']}: {entry['mode']}")
+        if entry.get("mode"):
+            if entry["mode"] not in ("analog", "alternate"):
+                raise ValueError(f"Invalid mode for {entry['name']}: {entry['mode']}")
 
     if entry.get("bias"):
         if entry["bias"] not in ("disable", "pull-up", "pull-down"):
@@ -418,11 +418,7 @@ def get_mcu_signals(cube_path, gpio_ip_afs):
                 if signal_name is None:
                     continue
 
-                if re.match(r"^ADC(?:\d+)?_IN[NP]?\d+$", signal_name) or re.match(
-                    r"^DAC(?:\d+)?_OUT\d+$", signal_name
-                ):
-                    found_afs = [None]
-                elif signal_name in pin_afs:
+                if signal_name in pin_afs:
                     pin_af = pin_afs[signal_name]
                     if not isinstance(pin_af, list):
                         pin_af = [pin_af]
@@ -430,8 +426,9 @@ def get_mcu_signals(cube_path, gpio_ip_afs):
                 # STM32F1: assume NO_REMAP (af=0) if signal is not listed in pin_afs
                 elif family == "STM32F1":
                     found_afs = [0]
+                # Non STM32F1: No alternate function found, mode is analog
                 else:
-                    continue
+                    found_afs = [None]
 
                 for af in found_afs:
                     pin_signals.append({"name": signal_name, "af": af})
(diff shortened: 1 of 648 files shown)


scripts/genpinctrl/genpinctrl.py
```diff
@@ -429,7 +425,7 @@ def get_mcu_signals(cube_path, gpio_ip_afs):
                 elif family == "STM32F1":
                     found_afs = [0]
                 else:
-                    continue
+                    found_afs = [None]
```
gmarull, line 428: Would probably be good to emphasize that this `else` is for `analog` mode, it's less clear now.

scripts/genpinctrl/genpinctrl.py
```diff
@@ -505,6 +501,12 @@ def main(cube_path, output):
                         if af["name"] not in entries:
                             entries[af["name"]] = list()
 
+                        if family != "STM32F1":
```
gmarull, line 504: `mode` from config file should take precedence over inferred mode. So I think we should have:

```python
if not af.get("mode") and family != "STM32F1":
```

scripts/genpinctrl/stm32-pinctrl-config.yaml
```diff
@@ -11,10 +11,12 @@
 #     be as precise as possible. Note that it needs to be escaped here in the
 #     configuration file.
 #
-#   - mode (mandatory): Mode setting (analog, alternate). Mode needs to
+#   - mode (optional): Mode setting (analog, alternate). Mode needs to
 #     be set according to the following rules:
 #       * Pin operates in analog configuration: analog
 #       * Pin operates in alternate function configuration: alternate
+#     In default cases, mode could be infered from inputs files and could be
```
gmarull, line 18: In default `case`, mode `can` be `inferred` from `CubeMX configuration files` and `can` be omitted.

gmarull (changes requested):
Added a few comments.

scripts/genpinctrl/genpinctrl.py
erwango: Added a comment

scripts/genpinctrl/genpinctrl.py
erwango: Indeed, fixed

scripts/genpinctrl/stm32-pinctrl-config.yaml
erwango: Fixed

gmarull (approved):

galak (approved):

galak:
@erwango looks like this needs rebasing.

erwango:
DNM as I've found late bug

erwango:
Late bug is the following:
This change doesn't play nice when signal is defined by group, such as:
```
- name: USB_OTG_HS
  match: "^USB_OTG_HS_(?:DM)?(?:DP)?(?:SOF)?(?:ID)?(?:VBUS)?$"
```
and if one of them the inferred mode differs from the other's.

erwango:
@gmarull I had to do some changes due to late bug discovery, can you review again ?

scripts/genpinctrl/genpinctrl.py
```diff
@@ -507,13 +504,24 @@ def main(cube_path, output):
                         if af["name"] not in entries:
                             entries[af["name"]] = list()
 
+                        # Define the signal mode using, by priority order:
+                        # 1- the config mode (ie: "af["mode"]")
+                        # 2- the inferred mode (an alternate function was found)
+                        if af.get("mode") is not None or family == "STM32F1":
```
gmarull, line 510: `if af.get("mode") or family == "STM32F1"` can be used (no need to check explicitely for None)

erwango: Looks better indeed. Fixed

scripts/genpinctrl/genpinctrl.py
```diff
@@ -507,13 +504,24 @@ def main(cube_path, output):
                         if af["name"] not in entries:
                             entries[af["name"]] = list()
 
+                        # Define the signal mode using, by priority order:
+                        # 1- the config mode (ie: "af["mode"]")
+                        # 2- the inferred mode (an alternate function was found)
+                        if af.get("mode") or family == "STM32F1":
+                            signal["mode"] = af["mode"]
+                        else:
```
gmarull, line 512: Actually, can this be compressed into a single if/elif/else block?

erwango: Sure, but wanted to make it clear.

gmarull:
LGTM

galak:
@erwango ready to merge?

erwango:
@galak, yes, this is fine now.

PR #85, opened 2020-12-15
scripts: serie_update:  minor fixes to be able to execute the script on ubuntu+zsh

str4t0m:
With my setup(ubuntu+zsh) the script crashes, when subprocess.check_call()
parameters are passed as list and shell=True is set,
or if subprocess.check_call() parameters are passed as string
and shell=True is not set. These changes fix that.

I did not yet have a chance to test if the changes break anything on Windows.

### DIFF
--- a/scripts/serie_update.py
+++ b/scripts/serie_update.py
@@ -30,6 +30,7 @@ def os_cmd(cmd, cwd=None, shell=False):
     Args:
         cmd: string command to execute.
         cwd: directory where to run command
+        shell: boolean to enable command interpretation by the shell
 
     Returns:
         return the returncode of the command after execution.
@@ -137,7 +138,8 @@ class Stm32SerieUpdate:
         # with the most recent one created being the last entry.
         os_cmd(("git", "checkout", "master"), cwd=self.stm32cube_serie_path)
         self.version_tag = subprocess.check_output(
-            "git tag -l", cwd=self.stm32cube_serie_path
+            ("git", "tag", "-l"),
+            cwd=self.stm32cube_serie_path
         ).splitlines()
         self.version_tag = [x.decode("utf-8") for x in self.version_tag]
         # Set latest version
@@ -265,11 +267,12 @@ class Stm32SerieUpdate:
         )
         os_cmd(
             (
-                "cp",
-                "-r",
-                str(stm32cube_drivers_src_path) + "/*.*",
-                str(temp_drivers_src_path),
-            )
+                "cp " +
+                "-r " +
+                str(stm32cube_drivers_src_path) + "/*.* " +
+                str(temp_drivers_src_path)
+            ),
+            shell=True,
         )
 
     def build_from_current_cube_version(self):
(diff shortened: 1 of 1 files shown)


ABOSTM (approved):
I tested this PR on windows, and it is working fine.
LGTM

str4t0m:
Thanks for testing this on windows.

scripts/serie_update.py
```diff
@@ -30,6 +30,7 @@ def os_cmd(cmd, cwd=None, shell=False):
     Args:
         cmd: string command to execute.
         cwd: directory where to run command
+        shell: boolean to enable command interpretation by the shell
```
erwango, line 33: Any impact on ./update_stm32_package.py ?
Unless this is totally transparent to users in any case, please update scripts/README.rst ?

erwango (changes requested):

scripts/serie_update.py
str4t0m: The function was just missing the arguments doc string, so I decided to include it in this commit.
Functionally nothing changed, can I leave it as it is?

erwango (approved):

scripts/serie_update.py
```diff
@@ -264,12 +265,9 @@ def extract_source(self):
...
         os_cmd(
-            (
-                "cp",
-                "-r",
-                str(stm32cube_drivers_src_path) + "/*.*",
-                str(temp_drivers_src_path),
-            )
+            "cp -r " + str(stm32cube_drivers_src_path) + "/*.* "
+            + str(temp_drivers_src_path),
+            shell=True,
```
gmarull, line 270: I understand that this is in fact the only change required, others are cosmetic and not strictly required (should go into a separate commit)

scripts/serie_update.py
```diff
@@ -355,15 +345,8 @@ def build_patch_from_current_zephyr_version(self):
...
-                (
-                    "git",
-                    "diff",
-                    "HEAD@{1}",
-                    "--",
-                    str(hal_conf),
-                    ">>",
-                    str(hal_conf_patch),
-                ),
+                "git diff HEAD@{1} -- " + str(hal_conf)
```
gmarull, line 348: in general it is better to use multi-line strings if they do not fit in one line, so I prefer the previous code in this case, others LGTM

gmarull (changes requested):
Main request: split into 2 commits (actual fix and cosmetic changes)

scripts/serie_update.py
str4t0m: Unfortunately, except the comment in line 33, this is the minimal set of changes to get it running.
The arguments are treated differently on unix systems than on windows.

[Popen Constructor](https://docs.python.org/2/library/subprocess.html#popen-constructor):
_"On Unix with shell=True,   . If args is a sequence, the first item specifies the command string, and any additional items will be treated as additional arguments to the shell itself"_

In this case "git" is executed without arguments, and the rest of the sequence are arguments passed to the shell.
I am afraid that there is no way on unix systems to avoid passing the whole command as a string for commands that need "shell=True". But if you know any better solution that could work I will try it.

erwango: @gmarull I confirm this is working. Is that blocking ?

gmarull (commented):
I see the problem, but instead of merging long strings, better do:
```
a = (
   "b" +
   "c" +
   ...
   "z"
)
```

when the command is too long. That is, replace `,` with `+`. Anyway I think that current script deserves a re-write, so I won't block.

str4t0m:
I have reformatted it now as suggested.

erwango:
@galak Can we merge this one ?

PR #87, opened 2021-01-12
dts: add HRTIM pinctrl

martinjaeger:
Add options for high-resolution timer HRTIM to pinctrl-config and generate new pinctrl DTS files.

I know that the HRTIM peripheral is not yet supported by Zephyr, but I'm working on a driver and would like to use the pinctrl properly already.

### DIFF
--- a/dts/st/f3/stm32f334c(4-6-8)tx-pinctrl.dtsi
+++ b/dts/st/f3/stm32f334c(4-6-8)tx-pinctrl.dtsi
@@ -108,6 +108,88 @@
 				pinmux = <STM32_PINMUX('A', 6, ANALOG)>;
 			};
 
+			/* HRTIM_CH */
+
+			hrtim1_cha1_pa8: hrtim1_cha1_pa8 {
+				pinmux = <STM32_PINMUX('A', 8, AF13)>;
+			};
+
+			hrtim1_cha2_pa9: hrtim1_cha2_pa9 {
+				pinmux = <STM32_PINMUX('A', 9, AF13)>;
+			};
+
+			hrtim1_chb1_pa10: hrtim1_chb1_pa10 {
+				pinmux = <STM32_PINMUX('A', 10, AF13)>;
+			};
+
+			hrtim1_chb2_pa11: hrtim1_chb2_pa11 {
+				pinmux = <STM32_PINMUX('A', 11, AF13)>;
+			};
+
+			hrtim1_chc1_pb12: hrtim1_chc1_pb12 {
+				pinmux = <STM32_PINMUX('B', 12, AF13)>;
+			};
+
+			hrtim1_chc2_pb13: hrtim1_chc2_pb13 {
+				pinmux = <STM32_PINMUX('B', 13, AF13)>;
+			};
+
+			hrtim1_chd1_pb14: hrtim1_chd1_pb14 {
+				pinmux = <STM32_PINMUX('B', 14, AF13)>;
+			};
+
+			hrtim1_chd2_pb15: hrtim1_chd2_pb15 {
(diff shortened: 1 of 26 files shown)


gmarull (changes requested):
@martinjaeger We usually add the config changes in one commit and the autogenerated content into another one to ease review process.

martinjaeger:
Ah, makes sense. Will fix that tomorrow.

gmarull (approved):

luizvilla:
Wonderful work @Martin!
We were looking for this. We'll implement and leave a comment here.

martinjaeger:
> Wonderful work @martinjaeger.
> We were looking for this. We'll implement and leave a comment here.

Great. Zephyr has feature freeze at the moment until v2.5 is released (scheduled for 12 February). After that we should probably get this PR in rather quickly.

In the meantime you can already add something like this to your board.dts as a workaround:
```
	soc {
		hrtim1: timers@40016800 {
			compatible = "st,stm32-timers";
			reg = <0x40016800 0x400>;
			clocks = <&rcc STM32_CLOCK_BUS_APB2 0x04000000>;
			interrupts = <67 0>, <68 0>, <69 0>, <70 0>, <71 0>, <72 0>, <73 0>, <74 0>;
			interrupt-names = "master", "tima", "timb", "timc", "timd", "time", "flt", "timf";
			status = "okay";
			label = "HRTIM_1";
		};

		/* required until hal_stm32 PR #87 is merged */
		pinctrl: pin-controller@48000000 {
			hrtim1_cha1_pa8: hrtim1_cha1_pa8 {
				pinmux = <STM32_PINMUX('A', 8, AF13)>;
			};

			hrtim1_cha2_pa9: hrtim1_cha2_pa9 {
				pinmux = <STM32_PINMUX('A', 9, AF13)>;
			};

			// ...
		};
	};
```

galak:
Is there a Zephyr side PR related to this?

martinjaeger:
> Is there a Zephyr side PR related to this?

Not yet. The driver is still work in progress.

It doesn't really make sense to use the HRTIM with the existing "simple" PWM API. The HRTIM is meant for motor driver and power converter applications with complementary PWM outputs. So we'll probably have to come up with a new API for that (half_bridge, sync_pwm, complementary_pwm or how ever it may be called).

erwango (approved):

erwango:
@martinjaeger Set as DNM waiting for a PR to use these in main Zephyr repo. Any objections ?

gmarull:
@erwango I think it is still fine to have definitions for peripherals that are currently not used in-tree. It can be helpful for people developing custom drivers.

martinjaeger:
> @erwango I think it is still fine to have definitions for peripherals that are currently not used in-tree. It can be helpful for people developing custom drivers.

I agree with @gmarull. The pinctrl driver is fully supported upstream and this PR only adds features to the pinctrl driver.

IMO we could even say that we generate pinctrl devicetree nodes that are defined in the original ST cube XML files in order to make out-of-tree driver development on top of pinctrl easier.

We don't need a dedicated commit upstream to update `west.yml`. But I think it would not harm anyone to just merge this here and get it in together with the next update.

erwango:
@gmarull, @martinjaeger Ok, it doesn't harm indeed and I don't have strong opinion otherwise.

> IMO we could even say that we generate all pinctrl devicetree nodes that are defined in the original ST cube XML files in order to make out-of-tree driver development on top of pinctrl easier.

I'm not particularly of favor of this as I prefer to have users and reviewers for each peripherals (as it seems to be that case here), and avoid generating blindly configurations that could be potentially broken. So I prefer this remains a case by case operation.

erwango:
@galak, this one is ready for merge.

PR #88, opened 2021-02-08
fix uart pull definitions

shlomow:
Use pull-up on inputs and remove it from outputs.
In the past, it was done intentionally since the
pinmux was configured before the uart driver. Now,
it is configured inside the uart driver using `pinctrl-0`
property.

Fixes: #https://github.com/zephyrproject-rtos/zephyr/issues/15349

### DIFF
--- a/dts/st/f0/stm32f030c6tx-pinctrl.dtsi
+++ b/dts/st/f0/stm32f030c6tx-pinctrl.dtsi
@@ -304,13 +304,11 @@
 
 			usart1_rts_pa1: usart1_rts_pa1 {
 				pinmux = <STM32_PINMUX('A', 1, AF1)>;
-				bias-pull-up;
 				drive-open-drain;
 			};
 
 			usart1_rts_pa12: usart1_rts_pa12 {
 				pinmux = <STM32_PINMUX('A', 12, AF1)>;
-				bias-pull-up;
 				drive-open-drain;
 			};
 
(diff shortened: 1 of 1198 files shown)


scripts/genpinctrl/genpinctrl.py
```diff
@@ -456,7 +456,7 @@ def main(data_path, output):
         config_f1 = yaml.load(f, Loader=yaml.Loader)
 
     env = Environment(
-        trim_blocks=True, lstrip_blocks=True, loader=FileSystemLoader(SCRIPT_DIR)
+        trim_blocks=True, lstrip_blocks=True, loader=FileSystemLoader(str(SCRIPT_DIR))
```
gmarull, line 459: Current codes is working well without (see last CI tests)

dts/st/f0/stm32f030c6tx-pinctrl.dtsi
```diff
@@ -304,54 +304,52 @@
...
 				pinmux = <STM32_PINMUX('A', 12, AF1)>;
-				bias-pull-up;
 				drive-open-drain;
 			};
 
 			/* UART_RX / USART_RX / LPUART_RX */
 
 			usart1_rx_pa3: usart1_rx_pa3 {
 				pinmux = <STM32_PINMUX('A', 3, AF1)>;
+				bias-pull-up;
```
gmarull, line 319: Pinmux is now configured straight after clock initialization, https://github.com/zephyrproject-rtos/zephyr/blob/master/drivers/serial/uart_stm32.c#L1318. I wonder what is the rationale behind this change, is it causing some sort of line toggling? Maybe driver should initialize pins even before clock init?

gmarull (changes requested):

dts/st/f0/stm32f030c6tx-pinctrl.dtsi
shlomow: When you are connected to external board, the line can be in undefined state when the external board is in reset state or during initialization. If there is no pull up, then some faulty characters can be inserted or overrun error may be occurred. Most hardware designs put buffers or external pull ups, but if it is not the case then the internal pull up should be set.

gmarull: If defaults do not work for a particular setup they can be overridden at board level, e.g.
```
&usart1_rx_pa3 {
    bias-pull-up;
};
```

gmarull:
Is this PR still relevant?

erwango:
No update since more than 6 months. Closing.
Please re-open if needed.

PR #92, opened 2021-03-02
dts: Add slew-rate properties to stm32f1 series's SPI & I2S clock pins

ycsin:
Unlike other STM32 SOC (eg: F4, F7), the clock pins of SPI & I2S of STM32F1 series don't have the slew-rate properties defined, causing at least the SPI to not work if not defined in devicetree specifically. This pull request is to fix this issue after being enlightened by @erwango in Slack.

However, I've only tested the SPI1 for a SPI-NOR flash on my STM32F103RET6 custom board. I2S and other SPI port are not tested.

### DIFF
--- a/dts/st/f1/stm32f100c(4-6)tx-pinctrl.dtsi
+++ b/dts/st/f1/stm32f100c(4-6)tx-pinctrl.dtsi
@@ -122,10 +122,12 @@
 
 			spi1_sck_master_pa5: spi1_sck_master_pa5 {
 				pinmux = <STM32F1_PINMUX('A', 5, ALTERNATE, NO_REMAP)>;
+				slew-rate = "max-speed-50mhz";
 			};
 
 			spi1_sck_master_pb3: spi1_sck_master_pb3 {
 				pinmux = <STM32F1_PINMUX('B', 3, ALTERNATE, REMAP_1)>;
+				slew-rate = "max-speed-50mhz";
 			};
 
 			/* SPI_SLAVE_MISO */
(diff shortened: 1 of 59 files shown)


erwango (changes requested):
I don't see the changes made to `scripts/genpinctrl/stm32f1-pinctrl-config.yaml` for file generation.
Can you provide these in a dedicated commit ?

gmarull:
I imagine that SPI will not work properly at high speeds if slew rate is not defined correctly, unless it is documented that maximum slew rate is always needed. I think that `pinctrl` should always provide the low-speed default when possible, and in case higher speeds are needed, override `pinctrl` in the board `dts`. It is important to note that high slew-rates are responsible for many EMI problems if circuits are not properly designed.

erwango:
> I imagine that SPI will not work properly at high speeds if slew rate is not defined correctly, unless it is documented that maximum slew rate is always needed. I think that `pinctrl` should always provide the low-speed default when possible, and in case higher speeds are needed, override `pinctrl` in the board `dts`. It is important to note that high slew-rates are responsible for many EMI problems if circuits are not properly designed.

This makes sense. However, in former c macros definitions, SCK definitions were forced to 50MHz (in F1 case), and for some reason we missed it during the conversion. This used to be the case for several years, and as of today, we didn't see anyone complaining and request to move to slower speed. From that point of view, I'd consider it is safe.

gmarull:
To be honest, I'd be more comfortable if it is properly tested, e.g. what happens to SPI using multiple combinations of speeds and slew rates. I've seen a few designs failing EMI tests because of these subtle things. I can provide some scope traces in the next few days to compare and see if it is really needed in all situations.

ycsin:
> I don't see the changes made to `scripts/genpinctrl/stm32f1-pinctrl-config.yaml` for file generation.
> Can you provide these in a dedicated commit ?

Thanks for pointing out, its fixed now.

ycsin:
@gmarull If that's true I suppose those pins in `stm32-pinctrl-config.yaml` with `slew-rate: very-high-speed` need to change to `low-speed` as well?

gmarull:
@ycsin some peripherals, e.g. FMC or ETH require high speeds for all (or most common) scenarios, some others like SPI require high speeds only if they operate in the high speed range. I also know of a case (F4 SPI) that high speed is required because of an errata, it could be the case of F1 I don't know. Some peripherals were ported as they were defined in the legacy pinctrl header files, it's probably time to review them and see if low-speed defaults are a better option.

PR #94, opened 2021-03-30
dts: Align -pinctrl.dtsi on STM32CubeMX 6.2.0

erwango:
Update pinctrl.dtsi files based on recent STM32CubeMX 6.2.0
delivery.


erwango:
^^ @str4t0m

gmarull (approved):

FRASTM (approved):

galak:
is there a zephyr side pr related to this?

str4t0m:
The low end Cortex-m0 socs only have 6 to 7 alternative function mappings, but all mappings defined in the IP block are exported.
So we have a few extra definitions for af8, af9, ... mappings, that we could get rid of.
I think I'll open an issue that we consider this the next time the script is refactored.

str4t0m:
Sure. I'll point at an example in a dedicated issue, as it is not really relevant for this pr.

PR #99, opened 2021-06-22
Scripts update improvements (pushed on main branch)

ABOSTM:
Scripts update improvements (Pushed on main branch):

- Raise error when ZEPHYR_BASE is not defined
- Fix path error when git is not yet initialized
- Request confirmation before cleaning

### DIFF
--- a/scripts/serie_update.py
+++ b/scripts/serie_update.py
@@ -117,23 +117,25 @@ class Stm32SerieUpdate:
         self.latest_version = ""
         self.latest_commit = ""
 
-    def clone_cube_repo(self):
-        """Clone or fetch a stm32 serie repo"""
+    def major_branch(self):
         # check whether master branch exist, otherwise use main branch
         master_branch_exist = subprocess.check_output(
-            "git ls-remote --heads origin master", cwd=self.stm32cube_serie_path
+            ("git", "ls-remote", "--heads", "origin", "master"),
+            cwd=self.stm32cube_serie_path,
         ).decode("utf-8")
         if master_branch_exist:
-            branch = "master"
+            return "master"
         else:
-            branch = "main"
-        logging.info("Branch used:" + branch)
+            return "main"
 
+    def clone_cube_repo(self):
+        """Clone or fetch a stm32 serie repo"""
         if self.stm32cube_serie_path.exists():
             logging.info("fetching repo " + str(self.stm32cube_serie_path))
             # if already exists, then just clean and fetch
             os_cmd(("git", "clean", "-fdx"), cwd=self.stm32cube_serie_path)
             os_cmd(("git", "fetch"), cwd=self.stm32cube_serie_path)
+            branch = self.major_branch()
             os_cmd(
                 ("git", "reset", "--hard", branch),
                 cwd=self.stm32cube_serie_path,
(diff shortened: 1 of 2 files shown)


erwango (approved):

ABOSTM:
@FRASTM, @gmarull, gentle reminder

scripts/update_stm32_package.py
```diff
@@ -87,6 +87,18 @@ def update_cubes():
 else:
     repo_path = Path(os.getenv("HOME")) / "STM32Cube_repo"
 
+if not os.getenv("ZEPHYR_BASE"):
+    raise Exception("ZEPHYR_BASE Not defined")
+
+if not args.noclean:
+    print("Do you want to clean downloaded repo (" + str(repo_path) + ") at the end of updates?")
+    res = input("(Enter y/n) ").lower()
+    while res not in {"y", "n"}:
```
gmarull, line 96:
```suggestion
    while res not in ("y", "n"):
```

scripts/update_stm32_package.py
```diff
@@ -87,6 +87,18 @@ def update_cubes():
 else:
     repo_path = Path(os.getenv("HOME")) / "STM32Cube_repo"
 
+if not os.getenv("ZEPHYR_BASE"):
+    raise Exception("ZEPHYR_BASE Not defined")
+
+if not args.noclean:
+    print("Do you want to clean downloaded repo (" + str(repo_path) + ") at the end of updates?")
+    res = input("(Enter y/n) ").lower()
```
gmarull, line 95: suggestion: add default option, so that `Enter` will just choose that option

scripts/update_stm32_package.py
```diff
@@ -87,6 +87,18 @@ def update_cubes():
...
+    while res not in {"y", "n"}:
+        res = input("(Enter y/n) ").lower()
+    if res=="n":
+        logging.info("Add option --noclean")
+        args.noclean=True
```
gmarull, line 100:
```suggestion
    if res == "n":
        logging.info("Add option --noclean")
        args.noclean = True
```

scripts/serie_update.py
```diff
@@ -143,6 +145,9 @@ def clone_cube_repo(self):
                 ("git", "clone", STM32_CUBE_REPO_BASE + self.serie + ".git"),
                 cwd=self.stm32cube_repo_path,
             )
+            branch = self.major_branch()
+
+        logging.info("Branch used: " + branch)
```
gmarull, line 150:
```suggestion
        logging.info("Branch used: %s", branch)
```

scripts/update_stm32_package.py
ABOSTM: done

scripts/update_stm32_package.py
ABOSTM: done

scripts/update_stm32_package.py
ABOSTM: @gmarull Thanks for your review.
I prefer to keep an explicit choice rather than a default one to avoid any mistakes.
(also it is most probable that I would chose  `clean ` as default whereas @erwango would prefer no clean as default :smiley:  )

gmarull (approved):

fpistm (approved):
LGTM

PR #107, opened 2021-07-13
Fix sdmmc signals

johnkjellberg:
Signals for SDMMC must be configured as pull-up for some SD cards. I had an older SanDisk Ultra that didn't work without pull-up. All other cards I have tested works both with and without pull up.


erwango (approved):

ABOSTM (approved):
LGTM

PR #111, opened 2021-07-28
stm32cube/wl: ll_utils: Consider HSEPRE when configuring PLL using HSE

erwango:
HSE prescaler value is not taken into account when computing
PLL frequency using HSE as input clock.
Fix this.

### DIFF
--- a/stm32cube/stm32wlxx/README
+++ b/stm32cube/stm32wlxx/README
@@ -50,4 +50,9 @@ Patch List:
     Impacted files:
      drivers/include/stm32wlxx_hal_conf.h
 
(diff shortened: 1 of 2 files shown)


ABOSTM (approved):
LGTM

FRASTM:
LGTM
What about stm32Cube**wb**xx which has the same HSEPRE divider ?

FRASTM (approved):

erwango:
> What about stm32Cubewbxx which has the same HSEPRE divider ?

Good catch. I'll have a check if same issue is present on WB and raise a ticket if required (though implementation in Zephyr may be delayed, as there is no need reported until now).

PR #117, opened 2021-09-17
stm32h7 adc driver oversampling ratio depends on ADC instance

FRASTM:
Setting the Oversampling ratio with the
LL_ADC_ConfigOverSamplingRatioShift and LL_ADC_GetOverSamplingRatio functions must consider the ADC
instance and IP version. Some ADC have a 9-bit OVSR value
from 1 to 1024, other like ADC3 have 3-bit OVSR value.
Bit field position is also specific.

The ratio parameter is supposed to be a value from 1 to 1024 or [2, 4, 8, 16, 32, 64, 128, 256] for ADC3.
Which is not the same as in functions LL_ADC_ConfigOverSamplingRatioShift and LL_ADC_GetOverSamplingRatio for other series (which is LL_ADC_OVS_RATIO_xx)

required to fix https://github.com/zephyrproject-rtos/zephyr/issues/38606

### DIFF
--- a/stm32cube/stm32h7xx/drivers/include/stm32h7xx_ll_adc.h
+++ b/stm32cube/stm32h7xx/drivers/include/stm32h7xx_ll_adc.h
@@ -6238,10 +6238,22 @@ __STATIC_INLINE uint32_t LL_ADC_GetOverSamplingDiscont(ADC_TypeDef *ADCx)
   *         ADC state:
   *         ADC must be disabled or enabled without conversion on going
   *         on either groups regular or injected.
+  * @note   Caution: Oversampling Ratio is dependent to ADC instance and IP version:
+  *         For STM32H72x/3x: ADC3 has 3-bit OVSR, and 9-bit OVSR for ADC1/ADC2
+  *         For Rest of STM32H7xxx All ADCs have 9-bit OVSR.
   * @rmtoll CFGR2    OVSS           LL_ADC_ConfigOverSamplingRatioShift\n
   *         CFGR2    OVSR           LL_ADC_ConfigOverSamplingRatioShift
   * @param  ADCx ADC instance
-  * @param  Ratio This parameter can be in the range from 1 to 1024.
+  * @param  Ratio This parameter can be in the range from 1 to 1024 or ...
+  *         one of the following values, in case of ADC3 of ADC_VER_V5_V90 :
+  *         @arg @ref LL_ADC_OVS_RATIO_2
+  *         @arg @ref LL_ADC_OVS_RATIO_4
+  *         @arg @ref LL_ADC_OVS_RATIO_8
+  *         @arg @ref LL_ADC_OVS_RATIO_16
+  *         @arg @ref LL_ADC_OVS_RATIO_32
+  *         @arg @ref LL_ADC_OVS_RATIO_64
+  *         @arg @ref LL_ADC_OVS_RATIO_128
+  *         @arg @ref LL_ADC_OVS_RATIO_256
   * @param  Shift This parameter can be one of the following values:
   *         @arg @ref LL_ADC_OVS_SHIFT_NONE
   *         @arg @ref LL_ADC_OVS_SHIFT_RIGHT_1
@@ -6259,19 +6271,48 @@ __STATIC_INLINE uint32_t LL_ADC_GetOverSamplingDiscont(ADC_TypeDef *ADCx)
   */
 __STATIC_INLINE void LL_ADC_ConfigOverSamplingRatioShift(ADC_TypeDef *ADCx, uint32_t Ratio, uint32_t Shift)
 {
-  MODIFY_REG(ADCx->CFGR2, (ADC_CFGR2_OVSS | ADC_CFGR2_OVSR), (Shift | (((Ratio - 1UL) << ADC_CFGR2_OVSR_Pos))));
+    /* ADC channels ratio shift oversampling */
+#if defined(ADC_VER_V5_V90)
+    if (ADCx == ADC3)
+    {
+        MODIFY_REG(ADCx->CFGR2, (ADC_CFGR2_OVSS | ADC3_CFGR2_OVSR), (Shift | Ratio));
+    } else
+#endif /* ADC_VER_V5_V90 */
+    {
+        MODIFY_REG(ADCx->CFGR2, (ADC_CFGR2_OVSS | ADC_CFGR2_OVSR), (Shift | (((Ratio - 1UL) << ADC_CFGR2_OVSR_Pos))));
+    }
 }
 
 /**
   * @brief  Get ADC oversampling ratio
   *        (impacting both ADC groups regular and injected)
+  * @note   Caution: Oversampling Ratio is dependent to ADC instance and IP version:
+  *         For STM32H72x/3x: ADC3 has 3-bit OVSR, and 9-bit OVSR for ADC1/ADC2
+  *         For Rest of STM32H7xxx All ADCs have 9-bit OVSR.
   * @rmtoll CFGR2    OVSR           LL_ADC_GetOverSamplingRatio
   * @param  ADCx ADC instance
-  * @retval Ratio This parameter can be in the from 1 to 1024.
+  * @retval Ratio This parameter can be a value from 1 to 1024 or ...
+  *         one of the following values, in case of ADC3 of ADC_VER_V5_V90 :
+  *         @arg @ref LL_ADC_OVS_RATIO_2
+  *         @arg @ref LL_ADC_OVS_RATIO_4
+  *         @arg @ref LL_ADC_OVS_RATIO_8
+  *         @arg @ref LL_ADC_OVS_RATIO_16
+  *         @arg @ref LL_ADC_OVS_RATIO_32
+  *         @arg @ref LL_ADC_OVS_RATIO_64
+  *         @arg @ref LL_ADC_OVS_RATIO_128
+  *         @arg @ref LL_ADC_OVS_RATIO_256
 */
 __STATIC_INLINE uint32_t LL_ADC_GetOverSamplingRatio(ADC_TypeDef *ADCx)
 {
-  return (((uint32_t)(READ_BIT(ADCx->CFGR2, ADC_CFGR2_OVSR)) + (1UL << ADC_CFGR2_OVSR_Pos)) >> ADC_CFGR2_OVSR_Pos);
+#if defined(ADC_VER_V5_V90)
+    if (ADCx == ADC3)
+    {
+        return (uint32_t)(READ_BIT(ADCx->CFGR2, ADC3_CFGR2_OVSR));
+    } else
+#endif /* ADC_VER_V5_V90 */
+    {
+        return (((uint32_t)(READ_BIT(ADCx->CFGR2, ADC_CFGR2_OVSR)) + (1UL << ADC_CFGR2_OVSR_Pos)) >> ADC_CFGR2_OVSR_Pos);
+    }
 }
 
 /**
--- a/stm32cube/stm32h7xx/README
+++ b/stm32cube/stm32h7xx/README
@@ -36,6 +36,12 @@ License Link:
 
 Patch List:
 
+   *LL_ADC_ConfigOverSamplingRatioShift
+    -ADC3 oversampling Ratio incorrectly set
+    Impacted files:
+     drivers/include/stm32h7xx_ll_adc.h
+    ST Bug tracker ID: 113809
+
    *Changes from official delivery:
     -dos2unix applied
     -trailing white spaces removed


FRASTM:
add the change for the `LL_ADC_GetOverSamplingRatio()` function

stm32cube/stm32h7xx/drivers/include/stm32h7xx_ll_adc.h
```diff
@@ -6238,10 +6238,14 @@ __STATIC_INLINE uint32_t LL_ADC_GetOverSamplingDiscont(ADC_TypeDef *ADCx)
...
   *         on either groups regular or injected.
+  * @note   Caution: Oversampling Ratio is dependent to ADC instance and IP version:
+  *         For STM32H72x/3x: ADC3 has 3-bit OVSR, and 9-bit OVSR for ADC1/ADC2
+  *         For Rest of STM32H7xxx All ADCs have 9-bit OVSR.
   * @rmtoll CFGR2    OVSS           LL_ADC_ConfigOverSamplingRatioShift\n
   *         CFGR2    OVSR           LL_ADC_ConfigOverSamplingRatioShift
   * @param  ADCx ADC instance
-  * @param  Ratio This parameter can be in the range from 1 to 1024.
+  * @param  Ratio This parameter can be in the range from 1 or 1024
+  *               or 2, 4, 8, 16, 32, 64, 128, 256 for ADC3.
```
heinwessels, line 6248: This function should accept the 3-bit value, and not the 10-bit value. For example, it should be able to accept the define [LL_ADC_OVS_RATIO_2](https://github.com/STMicroelectronics/STM32CubeH7/blob/5975bffae9358bc2b2890a35a203d940a395efef/Drivers/STM32H7xx_HAL_Driver/Inc/stm32h7xx_ll_adc.h#L1397-L1412).

These defines seems to be shifted to the correct location already. E.g:
```
LL_ADC_OVS_RATIO_256 = 28 = 0b0011100
```
I understand that this is drivers only used by Zephyr, and that Zephyr might not use `LL_ADC_OVS_RATIO_2`, but the fix for this [#177](https://github.com/STMicroelectronics/STM32CubeH7/issues/177) should be able to accept `LL_ADC_OVS_RATIO_2`, which will create a conflict.

FRASTM: If so, this is a change in the LL API
LL_ADC_ConfigOverSamplingRatioShift ratio is a uint32 value from 1 to 1024, including for ADC3 of the stm32h72x, the range is 2..256.

erwango (changes requested):
Add fix description in README (+ add bug internal reference)

FRASTM:
internal bug [113089]

stm32cube/stm32h7xx/drivers/include/stm32h7xx_ll_adc.h
```diff
@@ -6259,19 +6263,43 @@ __STATIC_INLINE uint32_t LL_ADC_GetOverSamplingDiscont(ADC_TypeDef *ADCx)
   */
 __STATIC_INLINE void LL_ADC_ConfigOverSamplingRatioShift(ADC_TypeDef *ADCx, uint32_t Ratio, uint32_t Shift)
 {
-  MODIFY_REG(ADCx->CFGR2, (ADC_CFGR2_OVSS | ADC_CFGR2_OVSR), (Shift | (((Ratio - 1UL) << ADC_CFGR2_OVSR_Pos))));
+#if defined(ADC_VER_V5_V90)
+    if (ADCx == ADC3)
+    {
+        /* ADC channels ratio shift oversampling */
+        MODIFY_REG(ADCx->CFGR2, (ADC_CFGR2_OVSS | ADC3_CFGR2_OVSR), (Shift | ((Ratio >> ADC3_CFGR2_OVSR_Pos))));
```
heinwessels, line 6270: I'm not sure if this correct. E.g. if a ratio of 256 is required then the value written to `OVSR` should be `0b111`, and for 128 it should be `0b110`. A simple bitshift/divide will not suffice, since the relationship exponential with `2^(n+1).

Also is the bitshift supposed to move the value to the correct position? Then why `>>` and not `<<`?

However, I'm AFK at the moment so I cannot verify the code.

stm32cube/stm32h7xx/drivers/include/stm32h7xx_ll_adc.h
```diff
@@ -6259,19 +6263,43 @@ __STATIC_INLINE uint32_t LL_ADC_GetOverSamplingDiscont(ADC_TypeDef *ADCx)
...
   * @retval Ratio This parameter can be in the from 1 to 1024.
+  *               or 2, 4, 8, 16, 32, 64, 128, 256 for ADC3.
 */
 __STATIC_INLINE uint32_t LL_ADC_GetOverSamplingRatio(ADC_TypeDef *ADCx)
 {
-  return (((uint32_t)(READ_BIT(ADCx->CFGR2, ADC_CFGR2_OVSR)) + (1UL << ADC_CFGR2_OVSR_Pos)) >> ADC_CFGR2_OVSR_Pos);
+#if defined(ADC_VER_V5_V90)
+    if (ADCx == ADC3)
+    {
+        return (uint32_t)(2 << (READ_BIT(ADCx->CFGR2, ADC3_CFGR2_OVSR) >> ADC3_CFGR2_OVSR_Pos));
```
heinwessels, line 6296: See comment above about how value stored in register is not what will be returned by this function.

heinwessels (changes requested):

stm32cube/stm32h7xx/drivers/include/stm32h7xx_ll_adc.h
FRASTM: 'Ratio' is given from the sequence->oversampling, with possible value of 0, 1, 2,  ... 8, meaning a shift of  2^ sequence->oversampling

For the stm32H7, the ADC 3 of ADC_VER_V5_V90 : (((Ratio -1UL) << ADC3_CFGR2_OVSR_Pos))
Ratio = sequence->oversampling = 8 (oversampling 256x), the OVSR[2:0] =  0b111
Ratio = sequence->oversampling = 6 (oversampling 64x), the OVSR[2:0] =   0b101
Ratio = sequence->oversampling = 4 (oversampling 16x), the OVSR[2:0] =   0b011

for other ADC : (((1UL << Ratio) - 1UL) << ADC_CFGR2_OVSR_Pos
Ratio = sequence->oversampling = 8 (oversampling 256x), the OSVR[9:0] =  255
Ratio = sequence->oversampling = 6 (oversampling 64x), the OSVR[9:0] =  63
Ratio = sequence->oversampling = 4 (oversampling 16x), the OSVR[9:0] =  15

FRASTM:
@ heinwessels changed to reflect your comment

FRASTM:
because the parameter Ratio is 1..8 or 9 or 10, it it aligned on other soc series function meaning an oversampling  of 2^Ratio.
+remove extra parenthesis

FRASTM:
internal bug tracking : [113809]

heinwessels (approved):

stm32cube/stm32h7xx/drivers/include/stm32h7xx_ll_adc.h
```diff
@@ -6259,19 +6263,44 @@ __STATIC_INLINE uint32_t LL_ADC_GetOverSamplingDiscont(ADC_TypeDef *ADCx)
   */
 __STATIC_INLINE void LL_ADC_ConfigOverSamplingRatioShift(ADC_TypeDef *ADCx, uint32_t Ratio, uint32_t Shift)
 {
-  MODIFY_REG(ADCx->CFGR2, (ADC_CFGR2_OVSS | ADC_CFGR2_OVSR), (Shift | (((Ratio - 1UL) << ADC_CFGR2_OVSR_Pos))));
+#if defined(ADC_VER_V5_V90)
```
ABOSTM, line 6266: I don't think STMicroelectronics developers will change the API (they will keep backward compatibility).
This mean they will keep ratio as range 1..1024  at least for H7 without ADC_VER_V5_V90 (and for ADC_VER_V5_V90 on ADC1/2)
So I would prefer to keep original code in the 'else'.
And in the 'if' (ADC_VER_V5_V90 on ADC3) I propose to do like other series (like G0, L4): use LL defines like LL_ADC_OVS_RATIO_128 as input parameter, thus no shift in LL .
Of course this imply to adapt stm32 zephyr driver accordingly.

ABOSTM (changes requested):

stm32cube/stm32h7xx/drivers/include/stm32h7xx_ll_adc.h
FRASTM: OK, so we decide _NOT_  to change the `LL_ADC_ConfigOverSamplingRatioShift() `for stm32H7
--> keeping the 'ratio' as a value from 1..1024 for ADC1/2 and from 1..256 for ADC3.
It means that the zephyr driver for stm32H7 will give the 2^oversampling (and no more the oversampling value)
--> changing the zephyr adc_stm32.c driver accordingly.

For the stm32H7, the ADC 3 of ADC_VER_V5_V90 : the OVSR[2:0] bitfield is 'Ratio' (1UL << sequence->oversampling)
Ratio = 256 (sequence->oversampling =8), oversampling by 256 ; OVSR[2:0] = 0b111
Ratio = 64 (sequence->oversampling = 6), oversampling by 64 ; OVSR[2:0] = 0b101
Ratio = 8 (sequence->oversampling = 3), oversampling by 8 ; OVSR[2:0] = 0b010

for other ADC: the OSVR[9:0] bitfield is ((1UL << Ratio) - 1UL) << ADC_CFGR2_OVSR_Pos
Ratio = 1024 (sequence->oversampling =10), oversampling by 1024 ; OSVR[9:0] = 1023
Ratio = 256 (sequence->oversampling =8), oversampling by 256 ; OSVR[9:0] = 255
Ratio = 64 (sequence->oversampling =6), oversampling by 64 ; OSVR[9:0] = 63
Ratio = 8 (sequence->oversampling =3) oversampling by 8 ; OSVR[9:0] = 7

FRASTM:
This change is to set the ratio shift oversampling for ADC3 adc instance (OVSR[2:0] bitfield)
to  accept ratio parameter as a value = 2,4,8,16,32,64,128,256.
Does not change for ADC with OSVR[9:0] bitfield.

FRASTM:
Remark: considering the LL function of the [STM32CubeU5 serie](https://github.com/STMicroelectronics/STM32CubeU5/blob/main/Drivers/STM32U5xx_HAL_Driver/Inc/stm32u5xx_ll_adc.h) 
It appears that the 'ratio' parameter, its value depends on the ADC instance. It could also be the cahnge to apply to this stm32H7 serie:
```
  * @param  Ratio This parameter can be in the range from 1 to 1024 in case of ADC1 or 2 instance.
  * @param  Ratio This parameter can be one of the following values, in case of ADC3 :
  *         @arg @ref LL_ADC_OVS_RATIO_2
  *         @arg @ref LL_ADC_OVS_RATIO_4
  *         @arg @ref LL_ADC_OVS_RATIO_8
  *         @arg @ref LL_ADC_OVS_RATIO_16
  *         @arg @ref LL_ADC_OVS_RATIO_32
  *         @arg @ref LL_ADC_OVS_RATIO_64
  *         @arg @ref LL_ADC_OVS_RATIO_128
  *         @arg @ref LL_ADC_OVS_RATIO_256
  * @param  Shift This parameter can be one of the following values:
  *         @arg @ref LL_ADC_OVS_SHIFT_NONE
  *         @arg @ref LL_ADC_OVS_SHIFT_RIGHT_1
  *         @arg @ref LL_ADC_OVS_SHIFT_RIGHT_2
  *         @arg @ref LL_ADC_OVS_SHIFT_RIGHT_3
  *         @arg @ref LL_ADC_OVS_SHIFT_RIGHT_4
  *         @arg @ref LL_ADC_OVS_SHIFT_RIGHT_5
  *         @arg @ref LL_ADC_OVS_SHIFT_RIGHT_6
  *         @arg @ref LL_ADC_OVS_SHIFT_RIGHT_7
  *         @arg @ref LL_ADC_OVS_SHIFT_RIGHT_8
  *         @arg @ref LL_ADC_OVS_SHIFT_RIGHT_9  (1)
  *         @arg @ref LL_ADC_OVS_SHIFT_RIGHT_10 (1)
  *         @arg @ref LL_ADC_OVS_SHIFT_RIGHT_11 (1)
  *
  *        (1): Only for ADC3 instance.
  * @retval None
  */
__STATIC_INLINE void LL_ADC_ConfigOverSamplingRatioShift(ADC_TypeDef *ADCx, uint32_t Ratio, uint32_t Shift)
{
  if (ADCx == ADC3)
  {
    MODIFY_REG(ADCx->CFGR2, (ADC_CFGR2_OVSS | ADC3_CFGR2_OVSR), (Shift | (((Ratio - 1UL) << ADC3_CFGR2_OVSR_Pos))));
  }
  else
  {
    MODIFY_REG(ADCx->CFGR2, (ADC_CFGR2_OVSS | ADC_CFGR2_OVSR), (Shift | Ratio));
  }
}
```

FRASTM:
This is the **stm32h7xx_ll_adc.h**  with the LL_ADC_ConfigOverSamplingRatioShift() function which is aligned on the  version for the stm32U5 (stm32u5_ll_adc.h).
The 'ratio' parameter depends on the ADC instance.

FRASTM:
No more useful as the zephyr ADC driver directly accesses to the register through the MODIFY_REG macro.

PR #118, opened 2021-09-23
stm32f1 gpio AF remap ll function without AFIO_MAPR_SWJ_CFG

FRASTM:
To remap or un-remap the different LL_GPIO_AF_EnableRemap_xxx
or LL_GPIO_AF_DisableRemap_xxx does not need to change the
SWJ_CFG bits of the MAPR register.


FRASTM:
internal bug: 114125

erwango (changes requested):
Modification could be seen as an API change and I'm not sure it could be accepted in Cube package.
Nacking until we can get a approval from Cube package maintainers.

FRASTM:
Instead change in the STM32CubeF1   API, the modification is done in the zephyr code : https://github.com/zephyrproject-rtos/zephyr/pull/38741

PR #119, opened 2021-10-22
stm32cube: u5: fix double promotion when calculating flash latency

XenuIsWatching:
The STM32U5 only has a single precision floating point processor. UTILS_SCALE3_LATENCY0_FREQ and UTILS_SCALE3_LATENCY2_FREQ were defined as double constants along with float_t mostly likely being a float. C implicit promotion rules will want to make floats into doubles very easily with the whole line getting promoted to doubles which can be inefficient.

Unless if I'm missing something here, I don't think it's a good idea to use `float_t` here. I've seen it used in other calculations in this module for performing a precise calculation where you'd want that level of control....but in this case all being done here is a comparison.

This is apart of trying to get `-Wdouble-promotion` added within Zephyr and these errors were discovered through an automated CI.

*I'm unsure of you're process over here for pull requests or issues, if there's an internal bug tracker you may want to add this to*

Zephyr Pull Request: https://github.com/zephyrproject-rtos/zephyr/pull/38958
CI Build Error (on 13/20): https://buildkite.com/zephyr/zephyr/builds/42989#d50e76fe-e820-499d-9f8e-ba160857de87

### DIFF
--- a/stm32cube/stm32u5xx/drivers/src/stm32u5xx_ll_utils.c
+++ b/stm32cube/stm32u5xx/drivers/src/stm32u5xx_ll_utils.c
@@ -66,9 +66,9 @@
 #define UTILS_SCALE2_LATENCY1_FREQ    (50000000U)      /*!< HCLK frequency to set FLASH latency 1 in power scale 2 */
 #define UTILS_SCALE2_LATENCY2_FREQ    (75000000U)      /*!< HCLK frequency to set FLASH latency 2 in power scale 2 */
 #define UTILS_SCALE2_LATENCY3_FREQ    (100000000U)     /*!< HCLK frequency to set FLASH latency 3 in power scale 2 */
-#define UTILS_SCALE3_LATENCY0_FREQ    (12.5000000)    /*!< HCLK frequency to set FLASH latency 0 in power scale 3 */
+#define UTILS_SCALE3_LATENCY0_FREQ    (12.5000000f)    /*!< HCLK frequency to set FLASH latency 0 in power scale 3 */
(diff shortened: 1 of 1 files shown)


stm32cube/stm32u5xx/drivers/src/stm32u5xx_ll_utils.c
```diff
@@ -66,9 +66,9 @@
 #define UTILS_SCALE2_LATENCY1_FREQ    (50000000U)      /*!< HCLK frequency to set FLASH latency 1 in power scale 2 */
 #define UTILS_SCALE2_LATENCY2_FREQ    (75000000U)      /*!< HCLK frequency to set FLASH latency 2 in power scale 2 */
 #define UTILS_SCALE2_LATENCY3_FREQ    (100000000U)     /*!< HCLK frequency to set FLASH latency 3 in power scale 2 */
-#define UTILS_SCALE3_LATENCY0_FREQ    (12.5000000)    /*!< HCLK frequency to set FLASH latency 0 in power scale 3 */
+#define UTILS_SCALE3_LATENCY0_FREQ    (12.5000000f)    /*!< HCLK frequency to set FLASH latency 0 in power scale 3 */
```
erwango, line 69: Looks like the '.' shouldn't be there at all.

erwango (commented):
@XenuIsWatching Thanks for raising this.
I wonder if the issue isn't much more simple and if we shouldn't just get rid of the unexpected . in frequency definitions.
Let me check this internally, and I'll come back to you

erwango:
Replaced by https://github.com/zephyrproject-rtos/hal_stm32/pull/130

PR #123, opened 2021-11-19
dts/st/f1: Review remap field generation

erwango:
In order to simplify work on stm32f1 pinmux driver, encode more information in the remap field.
To achieve this, we only need to add peripheral name in the remap value.
For instance:
```
-				pinmux = <STM32F1_PINMUX('B', 6, ALTERNATE, NO_REMAP)>;
+				pinmux = <STM32F1_PINMUX('B', 6, ALTERNATE, I2C1_REMAP0)>;
```
The remap values are directly extracted from the input .xml files ("RemapBlock Name").

`NO_REMAP` value is kept and now describe only the non "remap-able" devices.


gmarull (approved):

ABOSTM (approved):
LGTM

PR #126, opened 2021-12-14
dts: Provide analog configuration for each pin

erwango:
Update generation script to provide analog configuration for each pin.
This configuration could be used as low power pin configuration, when possible.

### DIFF
--- a/scripts/genpinctrl/stm32-pinctrl-config.yaml
+++ b/scripts/genpinctrl/stm32-pinctrl-config.yaml
@@ -10,6 +10,8 @@
 #     STM32 xml database pin configuration names. The regular expression should
 #     be as precise as possible. Note that it needs to be escaped here in the
 #     configuration file.
+#     Note: Specific "ANALOG" value allows generation of analog pins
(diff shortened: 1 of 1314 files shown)


scripts/genpinctrl/stm32-pinctrl-config.yaml
```diff
@@ -33,6 +33,11 @@
 #
 
 ---
+- name: Low Power pin configuration
+  match: "ANALOG"
+  mode: analog
+  variant: lp
```
gmarull, line 39: `lp` may be obviated (there's no other variant for the `ANALOG` match)

gmarull (approved):
LGTM, just one comment regarding the `lp` variant.

I'd also remove mentions to `low-power` here: https://github.com/zephyrproject-rtos/hal_stm32/blob/161c4b2a1b86b820c1f205e54dfbea3988195d40/scripts/genpinctrl/stm32-pinctrl-config.yaml#L30-L32

ABOSTM (approved):

PR #128, opened 2022-01-04
stm32H7 spi LL driver read address of the Tx/Rx reg. for DMA

FRASTM:
This patch adds the LL functions to get the address of SPI Tx/Rx reg.
Unlike other series, the stm32H7 has two registers for Tx and Rx
but no LL function is available to get their location when DMA requires
it for its peripheral address reg.

### DIFF
--- a/stm32cube/stm32h7xx/drivers/include/stm32h7xx_ll_spi.h
+++ b/stm32cube/stm32h7xx/drivers/include/stm32h7xx_ll_spi.h
@@ -2376,6 +2376,21 @@ __STATIC_INLINE uint32_t LL_SPI_IsEnabledDMAReq_TX(SPI_TypeDef *SPIx)
 }
 
 /**
+  * @brief  Get the data register address used for DMA transfer
+  * @rmtoll TXDR           TXDR            LL_SPI_DMA_GetRegAddr
+  * @param  SPIx SPI Instance
+  * @retval Address of data register
+  */
+__STATIC_INLINE uint32_t LL_SPI_DMA_GetTxRegAddr(SPI_TypeDef *SPIx)
+{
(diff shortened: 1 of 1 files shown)


stm32cube/stm32h7xx/drivers/include/stm32h7xx_ll_spi.h
```diff
@@ -2375,6 +2375,21 @@ __STATIC_INLINE uint32_t LL_SPI_IsEnabledDMAReq_TX(SPI_TypeDef *SPIx)
...
+  * @retval Address of data register
+  */
+__STATIC_INLINE uint32_t LL_SPI_DMA_GetTxRegAddr(SPI_TypeDef *SPIx)
+{
+  return (uint32_t) &(SPIx->TXDR);
+}
+__STATIC_INLINE uint32_t LL_SPI_DMA_GetRxRegAddr(SPI_TypeDef *SPIx)
+{
+  return (uint32_t) &(SPIx->RXDR);
+}
```
erwango, line 2391: Any idea if this will be integrated in Cube ?

erwango (changes requested):
Let's wait to see if this will be integrated in Cube first

stm32cube/stm32h7xx/drivers/include/stm32h7xx_ll_spi.h
FRASTM: internal bug tracking #120222

FRASTM: pallned for next stm32CubeH7 version > 1.9.1

erwango:
Available in H7 code since last release update

PR #129, opened 2022-01-27
stm32cube: stm32wl: Fix LL_SYSCFG_GetEXTISource

erwango:
This function was not returning the expected port.
Fix it


FRASTM (commented):
what about the LL_SYSCFG_**Set**EXTISource

FRASTM (commented):
the SYSCFG->EXTICR[Line & 0x03U]   gives the offset in the register bank
coming from the lower bits of the LL_SYSCFG_EXTI_LINEx  which is 0 or 1 or 2 or 3
there is no more than 4 registers

ABOSTM (approved):

FRASTM (approved):

PR #131, opened 2022-01-27
dts: provide uart/usart/lpuart de pins

GeorgeCGV:
Updated generation script to provide de (Driver Enable) pin configurations.
The driver enable pin is used for RS-485 and IO-Link communications with the STM32 USART.

There is an ongoing https://github.com/zephyrproject-rtos/zephyr/pull/40907 (RS-485 API PR) and #131 is required to provide RS-484 STM32 implementation.

### DIFF
--- a/scripts/genpinctrl/stm32-pinctrl-config.yaml
+++ b/scripts/genpinctrl/stm32-pinctrl-config.yaml
@@ -219,6 +219,10 @@
   drive: open-drain
   bias: pull-up
 
+- name: UART_DE / USART_DE / LPUART_DE
+  match: "^(?:LP)?US?ART\\d+_DE$"
+  drive: push-pull
+
 - name: UART_TX / USART_TX / LPUART_TX
   match: "^(?:LP)?US?ART\\d+_TX$"
   bias: pull-up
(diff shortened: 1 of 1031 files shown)


scripts/genpinctrl/stm32-pinctrl-config.yaml
```diff
@@ -219,6 +219,10 @@
   drive: open-drain
   bias: pull-up
 
+- name: UART_DE / USART_DE / LPUART_DE
```
erwango, line 222: Would you mind adding the same for F1 template ?

erwango (changes requested):
Thanks for pushing this.
Would you mind:
- also update F1 series (has a dedicated template)
- provide messages in your commits

Thanks!

scripts/genpinctrl/stm32-pinctrl-config.yaml
GeorgeCGV: The F1 and F2 SoCs don't seem to support/have RS-485 configuration.

erwango (approved):

FRASTM (approved):

PR #138, opened 2022-04-04
stm32cube: stm32f4xx: Use Legacy ethernet HAL

erwango:
A new ETH API was provided in last F4 HAL release.
Since Zephyr driver is not yet updated, don't use it and enable
what is now the Legacy API instead.


FRASTM (approved):
The stm32f4xx_hal_conf.h is changed from the stm32cube because of the zephyr. This could be mentionned here (through a comment) and/or in the ../modules/hal/stm32/stm32cube/stm32f4xx/README

PR #139, opened 2022-05-01
stm32cube: f4/h7: Use Zephyr Kconfig flag to control eth API

gleichdick:
Needed for zephyrproject-rtos/zephyr#45260


stm32cube/CMakeLists.txt
```diff
@@ -51,6 +51,10 @@ if(CONFIG_CLOCK_STM32_HSE_CLOCK)
   zephyr_compile_definitions( -DHSE_VALUE=${CONFIG_CLOCK_STM32_HSE_CLOCK} )
 endif()
 
+if(CONFIG_PTP_CLOCK_STM32_HAL)
+  zephyr_compile_definitions( -DHAL_ETH_USE_PTP)
+endif()
```
erwango, line 56: Please make it a dedicated commit as this is different from enabling new eth HAL API.

stm32cube/stm32f4xx/CMakeLists.txt
```diff
@@ -24,7 +24,7 @@ zephyr_library_sources_ifdef(CONFIG_USE_STM32_HAL_DMA drivers/src/stm32f4xx_hal_
 zephyr_library_sources_ifdef(CONFIG_USE_STM32_HAL_DMA2D drivers/src/stm32f4xx_hal_dma2d.c)
 zephyr_library_sources_ifdef(CONFIG_USE_STM32_HAL_DMA_EX drivers/src/stm32f4xx_hal_dma_ex.c)
 zephyr_library_sources_ifdef(CONFIG_USE_STM32_HAL_DSI drivers/src/stm32f4xx_hal_dsi.c)
-zephyr_library_sources_ifdef(CONFIG_USE_STM32_HAL_ETH drivers/src/Legacy/stm32f4xx_hal_eth.c)
+zephyr_library_sources_ifdef(CONFIG_USE_STM32_HAL_ETH drivers/src/stm32f4xx_hal_eth.c)
```
erwango, line 27: I'd prefer this is done using a dedicated new Kconfig flag.

erwango (changes requested):

gleichdick:
This is now selectable with the config option `ETH_STM32_USE_HAL_DRIVER_WITH_NEW_API`. Please note that the new driver offers callbacks for the PTP TX timestamp, so that poking in the tx descriptors is no longer necessary. That's why I chose to enable the PTP Hal driver when both PTP in general and using the new driver options are selected.

stm32cube/CMakeLists.txt
erwango: From what I can check this is only available with the new API, so dismissing this request.

stm32cube/CMakeLists.txt
```diff
@@ -51,6 +51,11 @@ if(CONFIG_CLOCK_STM32_HSE_CLOCK)
   zephyr_compile_definitions( -DHSE_VALUE=${CONFIG_CLOCK_STM32_HSE_CLOCK} )
 endif()
 
+if(CONFIG_PTP_CLOCK_STM32_HAL)
+  zephyr_compile_definitions_ifdef(CONFIG_ETH_STM32_HAL_API_V2
+    -DHAL_ETH_USE_PTP)
+endif()
```
erwango, line 57: Nit: I'd make it the other way round:
```
if(CONFIG_ETH_STM32_HAL_API_V2)
  zephyr_compile_definitions_ifdef(CONFIG_PTP_CLOCK_STM32_HAL
    -DHAL_ETH_USE_PTP)
endif()
```
1. This is more in line with other changes
2. Once new API is fully validated and the previous one is deprecated, we just need to remove the outer if() condition

stm32cube/stm32f4xx/CMakeLists.txt
```diff
@@ -24,7 +24,13 @@ zephyr_library_sources_ifdef(CONFIG_USE_STM32_HAL_DMA drivers/src/stm32f4xx_hal_
...
 zephyr_library_sources_ifdef(CONFIG_USE_STM32_HAL_DMA_EX drivers/src/stm32f4xx_hal_dma_ex.c)
 zephyr_library_sources_ifdef(CONFIG_USE_STM32_HAL_DSI drivers/src/stm32f4xx_hal_dsi.c)
-zephyr_library_sources_ifdef(CONFIG_USE_STM32_HAL_ETH drivers/src/Legacy/stm32f4xx_hal_eth.c)
+if(CONFIG_ETH_STM32_HAL_API_V2)
+  zephyr_compile_definitions(-DHAL_ETH_MODULE_ENABLED)
+  zephyr_library_sources(drivers/src/stm32f4xx_hal_eth.c)
+elseif(CONFIG_USE_STM32_HAL_ETH)
+  zephyr_compile_definitions(-DHAL_ETH_LEGACY_MODULE_ENABLED)
+  zephyr_library_sources(drivers/src/Legacy/stm32f4xx_hal_eth.c)
+endif()
```
erwango, line 33: As mentioned above, If possible, I'd prefer this is done with minimal impact on the existing code, so this could be removed with as much little changes once we'll get rid of the legacy API:

```suggestion
if(CONFIG_ETH_STM32_HAL_API_V2)
  zephyr_library_sources(drivers/src/stm32f4xx_hal_eth.c)
else
  zephyr_library_sources(drivers/src/Legacy/stm32f4xx_hal_eth.c)
endif()
```

stm32cube/stm32f4xx/drivers/include/stm32f4xx_hal_conf.h
```diff
@@ -44,8 +44,9 @@
 #define HAL_DCMI_MODULE_ENABLED
 #define HAL_DMA_MODULE_ENABLED
 #define HAL_DMA2D_MODULE_ENABLED
+/* ETH driver is chosen by CMake/Kconfig */
 /* #define HAL_ETH_MODULE_ENABLED */
-#define HAL_ETH_LEGACY_MODULE_ENABLED
+/* #define HAL_ETH_LEGACY_MODULE_ENABLED */
```
erwango, line 49: See above:
```suggestion
if(CONFIG_ETH_STM32_HAL_API_V2)
  #define HAL_ETH_MODULE_ENABLED
else
  #define HAL_ETH_LEGACY_MODULE_ENABLED
endif()
```

erwango (changes requested):
Please switch title to something more inline with project standards like:
`stm32cube: f4/h7: Use Zephyr Kconfig flag to control eth API`

stm32cube/stm32f4xx/drivers/include/stm32f4xx_hal_conf.h
gleichdick: CMake syntax in a header file? But I get it...

gleichdick:
I updated the PR according to the recommendations

erwango (approved):

PR #142, opened 2022-05-09
Add FMC bank address pinctrl

tmilkovic51:
`FMC_BA0` and `FMC_BA1` (bank address signals) are actually shared with `FMC_A14` and `FMC_A15` (meaning A14 and A15 pinctrl nodes can be referenced instead of BA0 and BA1 nodes), so feel free to decline this PR if you think having redundant pinctrl nodes (differing only in name) is unnecessary.


erwango:
@tmilkovic51 Thanks for proposal but we'll keep it simple for now. I'm closing this PR

PR #143, opened 2022-05-10
scripts: genpinctrl: use <zephyr/...> include prefix

gmarull:
Zephyr has prefixed all of its includes with <zephyr/...>. While the old
mode can still be used (CONFIG_LEGACY_INCLUDE_PATH) and is still enabled
by default, it's better to be prepared for its removal in the future.


erwango:
ok, but please rebase, as I've just merged octo spi pins.

gmarull:
@erwango done

erwango (approved):

PR #145, opened 2022-05-24
dts: stm32f1: Add remap information in pin configuration names

erwango:
Pin remap configurations may take two different values
For instance:
USART3_CTS PA13 matches both remap values 00 and 01.
    
In a peripheral pin remapping configuration, all pins associated to the
device must share the same remap value.
For instance, on USART3
-remap0 cfg uses PB10, PB11, PB12, PB13, PB14, all configured with remap0
-remap1 cfg uses PC10, PC11, PC12, PB13, PB14, all configured with remap1
    
If pinctrl driver detects one of the pins has a remap value different
from the whole set, it will generate an error.

Note:
I've chosen to include the remap information between signal an variant, as this is the most logical place

### DIFF
--- a/dts/st/f1/stm32f100c(4-6)tx-pinctrl.dtsi
+++ b/dts/st/f1/stm32f100c(4-6)tx-pinctrl.dtsi
@@ -219,7 +219,7 @@
 				drive-open-drain;
 			};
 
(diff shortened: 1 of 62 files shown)


gmarull (approved):

ABOSTM (approved):
LGTM

PR #148, opened 2022-09-16
Wrap define UNUSED with an ifndef

rideyourstyle:
To prevent the error `... error: "UNUSED" redefined ...` we have to wrap the define with an `#ifndef`.
We can not remove our define and think that others also have this problem and so can benefit from this change.

### DIFF
--- a/stm32cube/stm32f0xx/README
+++ b/stm32cube/stm32f0xx/README
@@ -47,4 +47,11 @@ Patch List:
     -Added stm32cube/stm32f0xx/drivers/include/stm32_assert.h
     -Removed unused stm32cube/stm32f0xx/drivers/include/stm32_assert_template.h
 
+   *Wrap define UNUSED with an ifndef
+     This will prevent a compiler warning for the case the UNUSED macro
+     is already defined elsewhere in the project
+     Impacted files:
+      drivers/include/stm32f0xx_hal_def.h
+     ST Internal Reference: 136825
+
    See release_note.html from STM32Cube
--- a/stm32cube/stm32f0xx/drivers/include/stm32f0xx_hal_def.h
+++ b/stm32cube/stm32f0xx/drivers/include/stm32f0xx_hal_def.h
@@ -55,7 +55,9 @@ typedef enum
 
 /* Exported macro ------------------------------------------------------------*/
 
-#define UNUSED(X) (void)X      /* To avoid gcc/g++ warnings */
+#ifndef UNUSED
+  #define UNUSED(X) (void)X      /* To avoid gcc/g++ warnings */
+#endif
 
 #define HAL_MAX_DELAY      0xFFFFFFFFU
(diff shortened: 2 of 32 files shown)


rideyourstyle:
I have seen that you @erwango are working very actively for this project. I could not enter any reviewers and wonder if this request will be viewed without a reviewer. Thanks

erwango (changes requested):
I'm fine with the change, but please also extend to other series as well (when not done already).
Also, please follow project contribution guidelines to format your pull request. [This presentation](https://www.youtube.com/watch?v=mnVNLM7ynMQ) could also help.

rideyourstyle:
All other defines in this file are HAL related. I see no reason to wrap them as well.

I changed my commits to match your guidelines. Hopefully I have considered everything.

erwango:
> All other defines in this file are HAL related.

I meant other stm32 series of course.

Meanwhile, I've checked internally and I need to confirm that we can integrate this patch. Let me explain:
Current repo is kind of a fork of official STM32Cube packages. In general we only integrate patches that could be integrated in official STM32Cube packages.
When patches are not integrated, we need to report them in current repo, each time a new cube package version is updated. So we need to restrict the number of such patches to the minimum.
I'll see what i can do and I'll keep you posted.
Meanwhile, one way to help would be to push the same patch in official [STM32F4 repo](https://github.com/STMicroelectronics/STM32CubeF4)

rideyourstyle:
Added the same PR to the STM32F4 repo:
https://github.com/STMicroelectronics/STM32CubeF4/pull/144

rideyourstyle:
And added the modification to all other stm32 series

erwango:
> And added the modification to all other stm32 series.

Great, thanks.

> Would it be necessary to create a pull request for each STM32CubeXX repository like for the STM32CubeF4?

This should not be required. Such change should be reported to other series internally (as long as the initial one is accepted).

rideyourstyle:
I have to change the Singed-of-by because of our company policy. At the same time I merged all changes in one commit, as it is desired in the zephyr repository

@erwango Is there any progress on this task?

erwango:
@rideyourstyle Since your change has been accepted by ST. I'm ok to merge it.
Tough, in order to track that the patch should be ported until we get Cube package updates that integrate the change, please update each package README file with the following in "Patch list" section:
```
   *Wrap define UNUSED with an ifndef
     <change summary>
     Impacted files:
      <Impacted file>
     ST Internal Reference: 136825
```

rideyourstyle:
@erwango Added patch entry to every README file.
And because of the different formats of the definition, my last commit contains not all `#define UNUSED...` (search-replace :-( ). Wrapped the missing ones also with the `#ifndef`

erwango (approved):

FRASTM (approved):

PR #150, opened 2022-09-19
pinctrl: fix duplicate entries

gmarull:
This PR fixes a couple of issues in the genpinctrl script that lead to duplicate nodes:

- Handling of duplicate remappable entries in some G0 files
- Skip pins with analog switch (Pxy_C) as they are not supported (H7)

Files have been regenerated, so we should no longer have duplicate entries.

### DIFF
--- a/scripts/genpinctrl/genpinctrl.py
+++ b/scripts/genpinctrl/genpinctrl.py
@@ -416,6 +416,10 @@ def get_mcu_signals(data_path, gpio_ip_afs):
             if family == "STM32G0" and pin_name in ("PA9", "PA10"):
                 continue
 
+            # skip pins with analog switch (Pxy_C) (not supported)
+            if pin_name.endswith("_C"):
+                continue
+
             # obtain pin port (A, B, ...) and number (0, 1, ...)
             m = re.search(r"^P([A-Z])(\d+).*$", pin_name)
             if not m:
--- a/scripts/genpinctrl/genpinctrl.py
+++ b/scripts/genpinctrl/genpinctrl.py
@@ -410,8 +410,13 @@ def get_mcu_signals(data_path, gpio_ip_afs):
             if pin.get("Type") != "I/O":
                 continue
 
-            # obtain pin port (A, B, ...) and number (0, 1, ...)
             pin_name = pin.get("Name")
+
+            # skip duplicate remappable entries in some G0 files
+            if family == "STM32G0" and pin_name in ("PA9", "PA10"):
+                continue
+
+            # obtain pin port (A, B, ...) and number (0, 1, ...)
             m = re.search(r"^P([A-Z])(\d+).*$", pin_name)
             if not m:
                 continue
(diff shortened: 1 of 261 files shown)


erwango (changes requested):
Thanks @gmarull for tackling this.
One nitpick: Please update used STM32_open_pin_data version `dts/README.rst`

gmarull:
> Thanks @gmarull for tackling this.
> One nitpick: Please update used STM32_open_pin_data version `dts/README.rst`

Version has not changed (readme is actually autogenerated)

erwango (approved):

mbolivar-nordic (approved):

mbolivar-nordic:
thanks for your help, @gmarull!

mbolivar-nordic:
Something is wrong:

```sh
$ west build -b stm32g0316_disco samples/hello_world
[...]
-- Found BOARD.dts: /home/mbolivar/zp/zephyr/boards/arm/stm32g0316_disco/stm32g0316_disco.dts
devicetree error: /soc/serial@40013800: undefined node label 'usart1_tx_pa9'
```

gmarull:
> Something is wrong:
> 
> ```shell
> $ west build -b stm32g0316_disco samples/hello_world
> [...]
> -- Found BOARD.dts: /home/mbolivar/zp/zephyr/boards/arm/stm32g0316_disco/stm32g0316_disco.dts
> devicetree error: /soc/serial@40013800: undefined node label 'usart1_tx_pa9'
> ```

I see we actually support pin remaps! Let me correct this PR then.

scripts/genpinctrl/genpinctrl.py
```diff
@@ -410,8 +410,13 @@ def get_mcu_signals(data_path, gpio_ip_afs):
             if pin.get("Type") != "I/O":
                 continue
 
-            # obtain pin port (A, B, ...) and number (0, 1, ...)
             pin_name = pin.get("Name")
+
+            # skip duplicate remappable entries in some G0 files
+            if family == "STM32G0" and pin_name in ("PA9", "PA10"):
+                continue
```
erwango, line 417: I understand that for such specific conditions, ad hoc code saves us from great complexity.
But thinking about potential upcoming series support, is there a way to generate a warning when duplicate nodes are seen ?
Not a blocker as I understand that zephyr dtlib will report issues anyway.

erwango (approved):

scripts/genpinctrl/genpinctrl.py
mbolivar-nordic: > Not a blocker as I understand that zephyr dtlib will report issues anyway.

I'll merge with your approval, then, thank you. I would like to get the fix in for v3.2.

gmarull: > I understand that for such specific conditions, ad hoc code saves us from great complexity. But thinking about potential upcoming series support, is there a way to generate a warning when duplicate nodes are seen ? Not a blocker as I understand that zephyr dtlib will report issues anyway.

They are not strictly duplicate, but rather redundant entries because G0 has the "single" and "remappable" entries, which basically encode the same information. And it's not consistent with F0 series (where pin remap functionality is also available). It would be good to clarify why XML is structured like that before doing any changes.

erwango: > It would be good to clarify why XML is structured like that before doing any changes.

Ok, I'll investigate

PR #151, opened 2022-09-26
scripts: genpinctrl: handle pins with analog switch

gmarull:
This patch adds support for handling pins with analog switch (found in
H7 series). This feature was removed in a previous patch because it was
generating duplicate nodes. However, it removed some valid combinations.
This patch adds them back but with proper suffix (_c) avoiding duplicate
entries.

### DIFF
--- a/dts/st/h7/stm32h723zgtx-pinctrl.dtsi
+++ b/dts/st/h7/stm32h723zgtx-pinctrl.dtsi
@@ -196,6 +196,18 @@
 				pinmux = <STM32_PINMUX('C', 1, ANALOG)>;
 			};
 
+			/omit-if-no-ref/ adc3_inn1_pc2_c: adc3_inn1_pc2_c {
+				pinmux = <STM32_PINMUX('C', 2, ANALOG)>;
+			};
+
+			/omit-if-no-ref/ adc3_inp0_pc2_c: adc3_inp0_pc2_c {
+				pinmux = <STM32_PINMUX('C', 2, ANALOG)>;
+			};
+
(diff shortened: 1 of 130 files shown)


erwango:
@GeorgeCGV

GeorgeCGV (approved):
LGTM

ps: thanks, that was quick

fabiobaltieri (approved):

dts/st/h7/stm32h723zgtx-pinctrl.dtsi
```diff
@@ -2045,6 +2091,26 @@
 				slew-rate = "very-high-speed";
 			};
 
+			/omit-if-no-ref/ octospim_p1_io2_pc2_c: octospim_p1_io2_pc2_c {
```
erwango, line 2094: Sorry, but I'm not sure to follow. If these nodes are not duplicate. Why do they need to be disambiguated ?

fabiobaltieri: I was wondering the same but then realized that the the pin name in the device datasheet is actually called PC2_C so I guess this makes sense.

gmarull: > I was wondering the same but then realized that the the pin name in the device datasheet is actually called PC2_C so I guess this makes sense.

yep, that's the point. previous file were actually wrong to my understading.

erwango: Ok, then this will require an update in release note.

erwango (approved):

PR #153, opened 2022-10-31
added the SAI pins to stm32-pinctrl

thanhduongvs:
This patch adds support for handling pins with Serial Audio Interface (SAI) signals.


erwango:
@thanhduongvs Would you mind rebasing, there were changes made recently to avoid duplicated nodes.
I'd like to be sure this is take into account.
Also github informs there are 125 lines deleted, I find this a little bit suspicious, but I didn't found where the lines are removed.

erwango:
Stale, closing.
Please reopen if required and address the comment

PR #154, opened 2022-11-23
lib: stm32wb: hci list do not disable irq when LST_remove_node will

FRASTM:
There is no need to disable the irq and backup/restore PRIMASK in the LST_remove_head and LST_remove_tail functions of the lib/stm32wb/hci/stm_list.c because the call to LST_remove_node() will do. If done twice a BUS FAULT occurs.

Fixes https://github.com/zephyrproject-rtos/zephyr/issues/50655


FRASTM:
internal ticket : 139525

PR #158, opened 2023-01-10
PTP ETH_STM32_HAL_API_V2

Desvauxm-st:
drivers: STM32H7: ethernet: PTP clock doesn’t works

fix to the V2 HAL API to get PTP to work
In the HAL_ETH_ReadData function where it checks for the last descriptor, I added a checked if the TSA bit was set in DESC1
If the TSA bit is set then have a peak at the context descriptor which should be the one after the last descriptor
If the CTXT bit is set in the context descriptor then extract the timestamps

### DIFF
--- a/stm32cube/stm32h7xx/README
+++ b/stm32cube/stm32h7xx/README
@@ -47,6 +47,16 @@ Patch List:
     - Impacted file: stm32h7xx_hal_eth.c
     - Internal reference: Not available. Will be fixed as part of a new eth hal implementation
 
+   *fix to the V2 HAL API to get PTP to work
+     In the HAL_ETH_ReadData function where it checks for the last descriptor, 
+     we added a checked if the TSA bit was set in DESC1
+     If the TSA bit is set then have a peak at the context descriptor which should be the one 
+     after the last descriptor
+     If the CTXT bit is set in the context descriptor then extract the timestamps
+     Impacted files:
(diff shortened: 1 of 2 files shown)


stm32cube/stm32h7xx/README
```diff
@@ -47,6 +47,15 @@ Patch List:
...
 
+   *fix to the V2 HAL API to get PTP to work
+     In the HAL_ETH_ReadData function where it checks for the last descriptor, 
+     we added a checked if the TSA bit was set in DESC1
+     If the TSA bit is set then have a peak at the context descriptor which should be the one 
+     after the last descriptor
+     If the CTXT bit is set in the context descriptor then extract the timestamps
+     Impacted files:
+     drivers/src/stm32h7xx_hal_eth.c
+
```
erwango, line 58: Please add     - Internal reference:

erwango (commented):
Last comment otherwise this is ok

chris1seto (commented):
I notice that the context descriptor is never added to the count of descriptors to be worked on by "ETH_UpdateDescriptor()". That function works to allocate new buffers to rebuild descriptors, and then it sets the OWN bit to return the descriptor to DMA. The context descriptor doesn't need a new buffer allocated, but it DOES need the OWN bit set. Where is that accomplished in the eth mac driver?

PR #100, opened 2021-06-23
assert: adding provision to enable asserts for all series

ikmdani:
These commits has changes necessary to make use of the present STM32Cube
hal & ll asserts for all series of stm32 platforms. Also the unused files has been
removed. Documentation updated.

PR #161, opened 2023-02-09
Support c0 series

benjaminbjornsson:
Add STM32C0 support:

STM32 HAL/LL
common_ll
pinctrl


erwango (commented):
I'll investigate on the duplicated node and will get back to you.
Otherwise looks fine

str4t0m (commented):
On STM32G0 there was the same issue regarding duplicate nodes, so you can reuse the fix in the `genpinctrl.py` script. Have a look at around line 423.
Regarding the question about a separate commit to remove the duplicates: I'd suggest to not squash manual changes into automatically generated content, but once you have changed the script as mentioned above this should no longer be necessary.

benjaminbjornsson:
> On STM32G0 there was the same issue regarding duplicate nodes, so you can reuse the fix in the `genpinctrl.py` script. Have a look at around line 423.
> Regarding the question about a separate commit to remove the duplicates: I'd suggest to not squash manual changes into automatically generated content, but once you have changed the script as mentioned above this should no longer be necessary.

Thanks, line 423 seems to be a good start. However, the current solution seems to depend on the order in which the duplicate nodes appear, which is reversed for C0 so the node with ANALOG mode is kept instead of the AF mode.

benjaminbjornsson:
I tried the same if-statement for C0 and it does remove duplicates, but the I2C1, TIM1 and USART1 nodes have analog pinmux mode.

benjaminbjornsson:
Okey, the problem seems to be when there's two nodes for PA9 or PA10 that have one node with no Variant attribute and one with Variant="PINREMAP" or Variant="PINREMAP_10_12". At least those are the once filtered on line 423 for G0, and also the once i had removed in the WIP commit.

erwango:
> Okey, the problem seems to be when there are two nodes for PA9 or PA10 that have one node with no Variant attribute and one with Variant="PINREMAP" or Variant="PINREMAP_10_12". At least those are the once filtered on line 423 for G0, and also the once i had removed in the WIP commit.

Are you investigating this ?
Otherwise, we can set it aside (keep the version with missing pins) and we'll have a look on our side when possible.

benjaminbjornsson:
> > Okey, the problem seems to be when there are two nodes for PA9 or PA10 that have one node with no Variant attribute and one with Variant="PINREMAP" or Variant="PINREMAP_10_12". At least those are the once filtered on line 423 for G0, and also the once i had removed in the WIP commit.
> 
> Are you investigating this ? Otherwise, we can set it aside (keep the version with missing pins) and we'll have a look on our side when possible.

Is there any reason not to generate with CubeMX 6.7.0 instead?

I've found that doing so removes the problem with some nodes having analog mode. It doesn't solve the problem of duplicates, but that could be solved in the same way as for G0.

erwango:
> I've found that doing so removes the problem with some nodes having analog mode. It doesn't solve the problem of duplicates, but that could be solved in the same way as for G0.

Ok, so likely this is an issue with STM32_open_pin_data repo which wasn't updated wrt latest CubeMX 6.7.0 database fixes.
I'll request an update.

benjaminbjornsson:
> > I've found that doing so removes the problem with some nodes having analog mode. It doesn't solve the problem of duplicates, but that could be solved in the same way as for G0.
> 
> Ok, so likely this is an issue with STM32_open_pin_data repo which wasn't updated wrt latest CubeMX 6.7.0 database fixes. I'll request an update.

What I can see, the 6.7.0 release is already available in the STM32_open_pin_data repo and we're currently using the 6.5.0 release. Should this be update?

benjaminbjornsson:
I opened another PR [164](https://github.com/zephyrproject-rtos/hal_stm32/pull/164) to use the newer revision of CubeMX.

benjaminbjornsson:
Updated again, sticking with the CubeMX release for now as suggested by [gautierg-st](https://github.com/gautierg-st) in [164](https://github.com/zephyrproject-rtos/hal_stm32/pull/164).

erwango (approved):

erwango:
Please rebase following merge of https://github.com/zephyrproject-rtos/hal_stm32/pull/162

benjaminbjornsson:
> Please rebase following merge of #162

Fixed

erwango (approved):

FRASTM (commented):
rename [stm32cube/stm32c0xx/drivers/include/stm32_assert_template.h](https://github.com/zephyrproject-rtos/hal_stm32/pull/161/files#diff-e31bda347262153c2df525f0c3daa27ddeeb8259daab9669d16c8fa19b67910f)

to stm32cube/stm32c0xx/drivers/include/stm32_assert.h

FRASTM:
> > rename [stm32cube/stm32c0xx/drivers/include/stm32_assert_template.h](https://github.com/zephyrproject-rtos/hal_stm32/pull/161/files#diff-e31bda347262153c2df525f0c3daa27ddeeb8259daab9669d16c8fa19b67910f)
> > to stm32cube/stm32c0xx/drivers/include/stm32_assert.h
> 
> Should it also just include stm32c0xx_hal_conf.h as for other series?

Yes, stm32c0xx_hal_conf.h is required  = renamed of stm32c0xx_hal_conf_template.h]

benjaminbjornsson:
> > > rename [stm32cube/stm32c0xx/drivers/include/stm32_assert_template.h](https://github.com/zephyrproject-rtos/hal_stm32/pull/161/files#diff-e31bda347262153c2df525f0c3daa27ddeeb8259daab9669d16c8fa19b67910f)
> > > to stm32cube/stm32c0xx/drivers/include/stm32_assert.h
> > 
> > 
> > Should it also just include stm32c0xx_hal_conf.h as for other series?
> 
> Yes, stm32c0xx_hal_conf.h is required = renamed of stm32c0xx_hal_conf_template.h

To clarify, I see that the `_assert.h` file for C0 is different from other series (e.g G0 where it only contains an include of stm32c0xx_hal_conf.h). Should this also be change to match other series, or only the name change?

FRASTM:
> > > > rename [stm32cube/stm32c0xx/drivers/include/stm32_assert_template.h](https://github.com/zephyrproject-rtos/hal_stm32/pull/161/files#diff-e31bda347262153c2df525f0c3daa27ddeeb8259daab9669d16c8fa19b67910f)
> > > > to stm32cube/stm32c0xx/drivers/include/stm32_assert.h
> > > 
> > > 
> > > Should it also just include stm32c0xx_hal_conf.h as for other series?
> > 
> > 
> > Yes, stm32c0xx_hal_conf.h is required = renamed of stm32c0xx_hal_conf_template.h
> 
> To clarify, I see that the `_assert.h` file for C0 is different from other series (e.g G0 where it only contains an include of stm32c0xx_hal_conf.h). Should this also be change to match other series, or only the name change?

--> just rename

benjaminbjornsson:
> > > > > rename [stm32cube/stm32c0xx/drivers/include/stm32_assert_template.h](https://github.com/zephyrproject-rtos/hal_stm32/pull/161/files#diff-e31bda347262153c2df525f0c3daa27ddeeb8259daab9669d16c8fa19b67910f)
> > > > > to stm32cube/stm32c0xx/drivers/include/stm32_assert.h
> > > > 
> > > > 
> > > > Should it also just include stm32c0xx_hal_conf.h as for other series?
> > > 
> > > 
> > > Yes, stm32c0xx_hal_conf.h is required = renamed of stm32c0xx_hal_conf_template.h
> > 
> > 
> > To clarify, I see that the `_assert.h` file for C0 is different from other series (e.g G0 where it only contains an include of stm32c0xx_hal_conf.h). Should this also be change to match other series, or only the name change?
> 
> --> just rename

Okey, fixed now :)

stm32cube/stm32c0xx/drivers/include/stm32_assert.h
```diff
@@ -0,0 +1,53 @@
+/**
```
erwango, line 1: Sorry but previous version was actually correct.
See https://github.com/zephyrproject-rtos/hal_stm32/pull/100

erwango (approved):

benjaminbjornsson:
@erwango, you were to quick :D There’s a copy-paste problem with the last change…

PR #164, opened 2023-03-03
dts: st: regenerate pinctrl files with CubeMX 6.7.0 release

benjaminbjornsson:
We're currently using the CubeMX 6.5.0 release which causes some problems when generating pinctrl for STM32C0. Updating to CubeMX 6.7.0 solves this problem.


erwango:
@gautierg-st please review

dts/st/f0/stm32f030r8tx-pinctrl.dtsi
```diff
@@ -496,7 +496,7 @@
 			};
 
 			/omit-if-no-ref/ tim3_ch3_pc8: tim3_ch3_pc8 {
-				pinmux = <STM32_PINMUX('C', 8, AF0)>;
+				pinmux = <STM32_PINMUX('C', 8, AF1)>;
```
gautierg-st, line 499: Previous code was correct. The datasheet agrees that it is AF0, and not AF1.
The issue has been raised on CubeMX teams side.

gautierg-st (changes requested):

benjaminbjornsson:
We'll wait for next release.

PR #172, opened 2023-05-04
Add TIM_BKIN signals

gmarull:
Add TIM_BKIN signals (used in power electronics applications). Note: updated to latest CubeMX before the changes.


erwango (changes requested):
Please add commit messages. I don't see valid reason it would be required in zephyr repo and optional in modules ;-)

Otherwise, LGTM

gmarull:
> Please add commit messages. I don't see valid reason it would be required in zephyr repo and optional in modules ;-)
> 
> Otherwise, LGTM

It's not needed in Zephyr, but in some out-of-tree projects (I think we have other signals like this). I'll open a PR in Zephyr.

erwango:
> It's not needed in Zephyr, but in some out-of-tree projects (I think we have other signals like this). I'll open a PR in Zephyr.

Well, it is mentioned quite explicitly here: 
https://docs.zephyrproject.org/latest/contribute/guidelines.html#commit-message-body

erwango (approved):

gautierg-st (approved):

PR #173, opened 2023-05-08
scripts: genpinctrl: add support for FSMC pins

douardda:
FSMC and FMC are very similar, but FSMC pins in STM32_open_pin_data are identified  as FSMC_xxx, so they are not included in generated pinctrl dts files.

This just modify the regex in the stm32-pinctrl-config.yaml file to match "FSMC" in addition to "FMC".

If accepted, dtsi files will have to be regenerated in a following PR.

### DIFF
--- a/scripts/genpinctrl/stm32-pinctrl-config.yaml
+++ b/scripts/genpinctrl/stm32-pinctrl-config.yaml
@@ -139,7 +139,7 @@
   match: "^FDCAN\\d+_TX$"
 
 - name: FMC
-  match: "^FMC_(?:NL|NADV|CLK|NBL[0-3]|A\\d+|D\\d+|NE[1-4]|NOE|NWE|NWAIT|NCE|INT|SDCLK|SDNWE|SDCKE[0-1]|SDNE[0-1]|SDNRAS|SDNCAS)$"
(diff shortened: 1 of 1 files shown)


douardda:
Any chance to have this PR reviewed?

erwango:
LGTM, but it's expected that you generate a new batch of -pinctrl.dtsi files by yourself (so we can see if the result is as expected).
Instructions are available in the repo README. Please take care to use the correct SHA1

erwango (changes requested):
Please generate new batch of files

douardda:
> 



> LGTM, but it's expected that you generate a new batch of -pinctrl.dtsi files by yourself (so we can see if the result is as expected).

Ok, it was not clear to me whether the newly generated dtsi files were expected with such a MR (so my comment "If accepted, dtsi files will have to be regenerated in a following PR") 

> Instructions are available in the repo README. Please take care to use the correct SHA1 and put this in a new commit.

I suppose "use the correct SHA1" is the sha1 mentioned in each `stm32cube/stm32xxx/README` file, right?
I realise I missed the README file in the `scripts/` directory, where all this is actually documented :-) (maybe a link from the top "repo README" file would help?)

erwango:
SHA1, is the one provided in `dts/README.rst`.
There are actually some lines in https://github.com/zephyrproject-rtos/hal_stm32#dtsi-files, but this can be improved and suggestions are welcome.

erwango:
I'm closing this PR due to inactivity. Please reopen if required.

PR #180, opened 2023-10-06
dts: Provide Serial-wire / JTAG pin description

erwango:
Provide Serial Wire / JTAG signal description.

Aim is to provide the ability to disable them (set to analog) in case is it possible and power consumption matters.
We don't provide a full extent description (speed, pull-down, ..) as there is no intention to support runtime change
of state from sleep to functional mode.


scripts/genpinctrl/stm32-pinctrl-config.yaml
```diff
@@ -267,3 +267,6 @@
 
 - name: USB
   match: "^USB_(?:DM)?(?:DP)?(?:NOE)?$"
+
+- name: JTAG PORT
+  match: ^(SYS|DEBUG)_(JTMS-SWDIO|JTCK-SWCLK|JTDI|JTDO(-TRACESWO|-SWO)?|(NJ)?JTRST)$
```
gautierg-st, line 272: On C0, the debug pins are "DEBUG_SWDIO" and "DEBUG_SWCLK".
On F0/G0/L0, they are "SYS_SWDIO" and "SYS_SWCLK".
Is it useful to add these patterns too?
Would the string
`^(SYS|DEBUG)_((JTMS-)?SWDIO|(JTCK-)?SWCLK|JTDI|JTDO(-TRACESWO|-SWO)?|(NJ)?JTRST)$`
work for these?

gautierg-st (commented):
Nothing is added for C0/F0/G0/L0. Is it intentional?

scripts/genpinctrl/stm32-pinctrl-config.yaml
erwango: Thanks, I'll have a check

gautierg-st (approved):

PR #181, opened 2023-11-16
Stm32h503  does not enable the driver if instance is not present

FRASTM:
The stm32h503 soc does not embed delayblock (DLYB) or SAI block or DCACHE instances.
Consequently the corresponding drivers should be included.


erwango (changes requested):
Please provide ST internal reference and document fix in README file

FRASTM:
Updating the stm32cube/stm32h5xx/README

erwango (approved):

gautierg-st (approved):

PR #184, opened 2023-11-24
scripts: serie_update: integration of changes in Git submodules of HAL & CMSIS Device drivers

ajarmouni-st:
STM32Cube HAL & CMSIS Device drivers are now included as Git submodules in the upstream STM32Cube repos. So we need to use additional or different Git commands to fetch them & integrate the changes correctly:

- To clone STM32Cube repos with the submodules, we add the "--recursive" option to "git clone".
- Instead of using "git reset --hard" to clean the working tree and switch to a given tag, which misses with submodules in various ways, we use "git checkout -f --recurse-submodules". "Using --recurse-submodules will update the content of all active submodules according to the commit recorded in the superproject. If local modifications in a submodule would be overwritten the checkout will fail unless -f is used." https://git-scm.com/docs/git-checkout

### DIFF
--- a/scripts/ble_library.py
+++ b/scripts/ble_library.py
@@ -136,9 +136,9 @@ def build_patch_from_current_zephyr_version(
 
     temp_source_lib_path = Path(temp_source_path / "lib")
 
-    # reset the STM32Cube repo to this current version
+    # Checkout the current Zephyr version of the STM32Cube repo
     os_cmd(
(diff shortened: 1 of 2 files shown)


erwango (changes requested):
LTGM, but please provide a bit more detailed summary of the changes in the commit message.

erwango (approved):

gautierg-st (approved):

PR #193, opened 2024-02-02
stm32cube: ll_spi: enhance interrupt register handling for stm32h5xx, h7xx, mp1xx, u5xx, wbaxx

ndrs-pst:
- add LL_SPI_EnableIT, LL_SPI_DisableIT
This update eliminates the need for individual calls to manipulate interrupt bits.

- use WRITE_REG to clear SPIx->IFCR
This adjustment reduces operations from read/write to write-only.

- add LL_SPI_ClearFlag
This update eliminates the need for individual calls to clear bits.

### DIFF
--- a/stm32cube/stm32h5xx/drivers/include/stm32h5xx_ll_spi.h
+++ b/stm32cube/stm32h5xx/drivers/include/stm32h5xx_ll_spi.h
@@ -1795,6 +1795,17 @@ __STATIC_INLINE uint32_t LL_SPI_GetRxFIFOPackingLevel(const SPI_TypeDef *SPIx)
 }
 
 /**
+  * @brief  Clear interrupt/status flag by setting corresponding bit in IFCR register
+  * @param  SPIx SPI Instance
+  * @param  ClearBits Clear bits
+  * @retval None
+  */
+__STATIC_INLINE void LL_SPI_ClearFlag(SPI_TypeDef *SPIx, uint32_t ClearBits)
+{
+  WRITE_REG(SPIx->IFCR, ClearBits);
+}
+
+/**
   * @brief  Clear End Of Transfer flag
   * @rmtoll IFCR         EOTC          LL_SPI_ClearFlag_EOT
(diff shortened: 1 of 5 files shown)


bryceschober:
@ndrs-pst I'm just in the "peanut gallery, but I suspect you could benefit from reviewing [the "important" section of the top-level README](https://github.com/zephyrproject-rtos/hal_stm32?tab=readme-ov-file#important).

> STM32Cube, as an external library, is not supposed to be modified...

There, they vaguely reference READMEs and patch lists, and ultimately I think they're referring to, for example for the STM32H5: 
<https://github.com/zephyrproject-rtos/hal_stm32/blob/main/stm32cube/stm32h5xx/README>

And in there, you'll find that they point to the actual upstream ST repository where these sources come from, for example for the STM32H5:
<https://github.com/STMicroelectronics/STM32CubeH5>, which actually contains your sources via a submodule that includes:
<https://github.com/STMicroelectronics/stm32h5xx_hal_driver>

I'm not sure whether this type of change would be sufficiently important to the Zephyr HAL maintainers to carry as a patchset on top of the upstream ST repository.

ndrs-pst:
@bryceschober Thank you for your advice!
I will try to open a pull request with the upstream ST repository.

erwango:
@ndrs-pst Thank you for looking into this, but Cube files in this repo are extracted from official STM32Cube repos.
We don't plan to diverge from these (for obvious maintenance reasons). We can merge temp fixes waiting for the integration in the official deliveries, but we won't merge any change when the direction is not approved by STM32Cube teams.
I hope you understand.

So, I'm closing this PR, but feel free to propose your changes first in official Cube repos:
https://github.com/STMicroelectronics/STM32CubeH7 and others.

Once you have a positive feedback, we can re-open this PR.

ndrs-pst:
@erwango Thank you for your guidance and for clarifying the process.
Initially, I considered opening another PR on the official Cube repos, but given your advice to propose changes directly,
I understand now that no further action is required on my part at this stage.

PR #204, opened 2024-05-06
dts: st: pinctrl.dtsi files including DCMI pins

CharlesDias:
Generate DCMI pins definitions for H7 family.

Related PR:

- https://github.com/zephyrproject-rtos/zephyr/pull/71462
- https://github.com/zephyrproject-rtos/zephyr/pull/71463

### DIFF
--- a/scripts/genpinctrl/stm32-pinctrl-config.yaml
+++ b/scripts/genpinctrl/stm32-pinctrl-config.yaml
@@ -52,6 +52,10 @@
 - name: DAC_OUT
   match: "^DAC(?:\\d+)?_OUT\\d+$"
 
+- name: DCMI
+  match: "^DCMI_(?:HSYNC|PIXCLK|VSYNC|D[0-7])$"
+  slew-rate: very-high-speed
+
 - name: ETH_COL
   match: "^ETH_COL$"
   slew-rate: very-high-speed
--- a/dts/st/f2/stm32f207i(c-e-f-g)hx-pinctrl.dtsi
+++ b/dts/st/f2/stm32f207i(c-e-f-g)hx-pinctrl.dtsi
@@ -834,6 +834,148 @@
 				pinmux = <STM32_PINMUX('A', 5, ANALOG)>;
 			};
 
+			/* DCMI */
+
+			/omit-if-no-ref/ dcmi_hsync_pa4: dcmi_hsync_pa4 {
+				pinmux = <STM32_PINMUX('A', 4, AF13)>;
+				slew-rate = "very-high-speed";
+			};
+
+			/omit-if-no-ref/ dcmi_pixclk_pa6: dcmi_pixclk_pa6 {
+				pinmux = <STM32_PINMUX('A', 6, AF13)>;
+				slew-rate = "very-high-speed";
+			};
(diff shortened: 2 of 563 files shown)


scripts/genpinctrl/stm32-pinctrl-config.yaml
```diff
@@ -52,6 +52,50 @@
...
+  match: "^DCMI_D5$"
+  slew-rate: very-high-speed
+
+- name: DCMI_D6
+  match: "^DCMI_D6$"
+  slew-rate: very-high-speed
+
+- name: DCMI_D7
+  match: "^DCMI_D7$"
+  slew-rate: very-high-speed
```
erwango, line 97: Could you use a regular expression ?
See for instance:
```
- name: FMC
  match: "^FMC_(?:NL|NADV|CLK|NBL[0-3]|A\\d+|D\\d+|NE[1-4]|NOE|NWE|NWAIT|NCE|INT|SDCLK|SDNWE|SDCKE[0-1]|SDNE[0-1]|SDNRAS|SDNCAS)$"
```
Aside from conciseness, this is slightly faster to generate.

erwango (changes requested):

scripts/genpinctrl/stm32-pinctrl-config.yaml
CharlesDias: Done

erwango:
Looks good to me but would it be possible to generate pins as well for other series?
I know driver is H7 specific for now (at least only validated on H7), but support might be extended in future. Having pins already available would be a good start.
This is the way we're proceeding in general.

CharlesDias:
Hi, @erwango.

I have two questions about the files generated after run the Python script:
1. Do I have to add also the new pinctrl files related to STM32H523, STM32H533, STM32H7R, and STM32H7s families? Those files are totally new.
2. I noticed that in `stm32h725iekx-pinctrl.dtsi`, `stm32h725igkx-pinctrl.dtsi`, `stm32h730ibkxq-pinctrl.dtsi`, and `stm32h735igkx-pinctrl.dtsi` files, the `adc3_inp16_ph5` was removed. Do I have to accept this modification? I'll check if something similar happens with other files.

erwango:
Thanks for raising this @CharlesDias.
This probably comes from the fact that you're generating from a latest version of open_pin_data repo.
2 options:
- Keep working with this commit, but then do it in 2 passes: first new batch from the new version w/o dcmi files, then add the new pinctrl, generate once again and provide the new signals in a 3 commit. In this case, don't forget to update commit used in https://github.com/zephyrproject-rtos/hal_stm32/blob/main/dts/README.rst (or maybe it is done automatically)
- Generate from SHA1 version provided in https://github.com/zephyrproject-rtos/hal_stm32/blob/main/dts/README.rst. And let the  open_pin_data update aside for now.

CharlesDias:
Hi, @erwango.

I choose the second option. ;-)

erwango:
This is fine for me now, but please update the first commit title to remove H7 reference, as pr is stm32 generic now.

erwango (approved):

PR #212, opened 2024-05-25
dts: Add DFSDM pins to pinctrl.dtsi files

charliegilliland:
I'm working on a dmic driver for Zephyr which will use DFSDM, so this adds those pins to the dtsi files and updates the config.yaml.

### DIFF
--- a/dts/st/f4/stm32f412c(e-g)ux-pinctrl.dtsi
+++ b/dts/st/f4/stm32f412c(e-g)ux-pinctrl.dtsi
@@ -238,6 +238,48 @@
 				pinmux = <STM32_PINMUX('B', 13, AF9)>;
 			};
 
+			/* DFSDM */
+
+			/omit-if-no-ref/ dfsdm1_datin1_pa4: dfsdm1_datin1_pa4 {
+				pinmux = <STM32_PINMUX('A', 4, AF8)>;
+				slew-rate = "very-high-speed";
+			};
+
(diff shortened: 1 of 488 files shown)


erwango (approved):

erwango:
@charliegilliland Would you min generating a new batch of -pinctrl.dtsi files based on latest top of this repo (05ee82ff431deefc32322cccf112b9b6aff4aec6) ?
We've just merged STM32H7RS pin control files and they would be impacted by you change.
Thanks

charliegilliland:
@erwango Sure thing, see my latest commits here.

erwango:
> @erwango Sure thing, see my latest commits here.

Thanks @charliegilliland but this will require a bit more git'nastic ;) :
We'll still the changes in 2 commits:
- script
- batch of new files
Also we don't take merge branches, so what you'd need to do is to edit the second of your initial commits and get the changes files here.
Depending on your state of tree the most simple might be to take the first commit on top of the new main, launch a new generation and make the new second commit from there.

erwango (changes requested):
See latest comment

charliegilliland:
@erwango OK, I did a rebase and re-generated the dtsi files. How does this look?
Thanks for your patience as I'm somewhat new to this workflow!

erwango (approved):

erwango:
> How does this look?

Perfect thanks.

> Thanks for your patience as I'm somewhat new to this workflow!

No issue, this is quite usual.

PR #220, opened 2024-06-27
stm32h7: fix TX timestamp check

kartben:
It looks like the original code was a bad copy-paste from H5 HAL. ETH_DMATXDESC_LS and ETH_DMATXDESC_TTSS are flags from the DESC0 register. This allows to effectively et the TX timestamp properly when there is one.

I wouldn't be surprised if more of such mix-ups of DESC0/1 vs., DESC2/3 would be to be found in the file but I didn't really review this.


erwango (changes requested):
I'm fine with the change, but please:
- Update README file (https://github.com/zephyrproject-rtos/hal_stm32/blob/d12e21e35d8d33c5fdf711d1a1c07f38c9295f70/stm32cube/stm32f7xx/README#L37)
- Open a ticket in official HAL repo to follow up the fix

erwango:
@marwaiehm-st Since @kartben is busy, would you mind taking over this change ?

marwaiehm-st:
> @marwaiehm-st Since @kartben is busy, would you mind taking over this change ?

Ok

erwango:
> This PR is replaced by #232

Thanks @marwaiehm-st.
@kartben we'll close current issue to make topic moving forward, if you don't mind.

PR #225, opened 2024-08-02
Add various enhancements to genpinctrl script

mathieuchopstm:
This PR adds three new features to the `genpinctrl` script to improve QoL, notably when introducing new series or regenerating pinctrl for new signals:
* graceful handling of non-Git pin data folders
  * If the pin data folder is not a git repository, script no longer crashes when generating the README file 
* support family filtering for (re)generation
  * In allowlist mode, pinctrl is generated only for specified families 
  * In forbidlist mode, pinctrl is generated for all families except those specified
  * __N.B.: when filtering is active, unlike in "default" mode, existing pinctrl files are not deleted__
* automatically detect the XML namespace from pin data files
  * Script will now attempt to detect the correct XMLNS from one file of the corpus
    * Files sourced from other places that the GitHub STMicroelectronics GitHub may use a different XML namespace
    * Before this modification, the script had to be manually edited
  * If automatic detection fails, the default `https://dummy.com` is still used
  * __N.B.: the script assumes that all files in the pin data use the same XMLNS__ (which should always be the case)

### DIFF
--- a/scripts/genpinctrl/genpinctrl.py
+++ b/scripts/genpinctrl/genpinctrl.py
@@ -79,6 +79,57 @@ PIN_MODS = [
 """Allowed pin modifiers"""
 
 
+class FamilyFilter():
+    def _prepare(self, filters, excluding: bool):
+        family_list = []
+
+        self.excluding = excluding
+
+        for filter in filters:
+            family_name = "STM32" + \
+                filter.upper().removeprefix("STM32")
+
+            family_list.append(family_name)
+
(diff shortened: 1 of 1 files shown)


erwango (approved):

erwango:
Merging despite the <3.9 failing tests, cf [python version ](https://github.com/zephyrproject-rtos/hal_stm32/commit/808602c497998f9e84f042ef90cf2010d69789dd)

PR #226, opened 2024-08-09
stm32h7xx_hal driver partial update to v1.11.3

FRASTM:
Some changes applied on the stm32h7xx_hal_driver
v1.11.3 are reported here. Until the stm32CubeH7 module is updated.

The version identified by the https://github.com/STMicroelectronics/STM32CubeH7  is v1.11.2 

So that stm32H7 serie is not updated 

However, the 
https://github.com/STMicroelectronics/stm32h7xx_hal_driver/tree/b2b4c45001f1b6b72ae449713607896fd6ca4c6b
is more a v1.11.3 

with some important changes like the one on the ospi hal driver which is put in this PR

### DIFF
--- a/stm32cube/stm32h7xx/drivers/src/stm32h7xx_hal_ospi.c
+++ b/stm32cube/stm32h7xx/drivers/src/stm32h7xx_hal_ospi.c
@@ -1597,7 +1597,7 @@ HAL_StatusTypeDef HAL_OSPI_Receive_DMA(OSPI_HandleTypeDef *hospi, uint8_t *pData
         }
 
         /* Enable the transmit MDMA Channel */
-        if (HAL_MDMA_Start_IT(hospi->hmdma, (uint32_t)pData, (uint32_t)&hospi->Instance->DR, hospi->XferSize, 1) == \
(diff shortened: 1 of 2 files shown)


erwango (changes requested):
v1.11.3 is [this branch](https://github.com/STMicroelectronics/stm32h7xx_hal_driver/commits/v1.11.3), but it doesn't contain the commit you mention. Also, picking up a commit (two from what I could tell), would require a clear update of the README

Since ST STM32Cube team is making use of rolling release in their HAL repo, we indeed need to adapt our cube update strategy, but we need to maintain clarity on our status.

I see 2 solutions:
-A] Cherry pick commits interesting for zephyr
-B] Point to the latest commit on top of the repo at a given time

A] allows to cherry pick only what we're interested in and be confident in our fork stability. Though, updates to tagged released could be trickier. And also, we need to display clearly our status vs original branch (ie clear update of the fork README)
With B],it will be more work, as we need to ensure the stability of all the commits we take (using non regression tests or commit per commit analysis), but updates should then be easier and status clearer at anytime.

I let you check what is the best in current case.

FRASTM:
Adding references in the README

erwango (approved):

marwaiehm-st (approved):

erwango (approved):

PR #227, opened 2024-08-22
zephyr: module.yml: Add missing module name

erwango:
Field `name` is missing in module.yml file.
Fix this.

### DIFF
--- a/zephyr/module.yml
+++ b/zephyr/module.yml
@@ -1,3 +1,4 @@
+name: hal_stm32
 build:
   cmake: .
   settings:


henrikbrixandersen (approved):

erwango:
@jhedberg FYI. Thanks for the heads up.

jhedberg (approved):

kartben (approved):

jhedberg:
@erwango btw, this changes the command used for fetching blobs from `west blobs fetch stm32` to `west blobs fetch hal_stm32`, so if you have any documentation with such references it may need to be updated.

erwango:
> @erwango btw, this changes the command used for fetching blobs from `west blobs fetch stm32` to `west blobs fetch hal_stm32`, so if you have any documentation with such references it may need to be updated.

Valuable information indeed. Thanks, I'll check this.

erwango:
> @erwango btw, this changes the command used for fetching blobs from `west blobs fetch stm32` to `west blobs fetch hal_stm32`, so if you have any documentation with such references it may need to be updated.

PR updated to keep the ability to keep the existing command `west blobs fetch stm32` functional.

jhedberg (changes requested):
> > @erwango btw, this changes the command used for fetching blobs from `west blobs fetch stm32` to `west blobs fetch hal_stm32`, so if you have any documentation with such references it may need to be updated.
> 
> PR updated to keep the ability to keep the existing command `west blobs fetch stm32` functional.

@erwango the reason I pointed out the difference was just to update documentation. I still think that `hal_stm32` is the correct name, since then it's consistent with how all other Zephyr HAL modules are defined and then it's also equal to the module name as it is defined in `west.yml`:

https://github.com/zephyrproject-rtos/zephyr/blob/143b14bb4fb779437fdeb8369e42fbb81db4a5fd/west.yml#L235

If you just call it `stm32` it's equally (if not more) likely that people will be confused when they are not able to fetch blobs the same way as they do for other modules. Furthermore, the name `hal_stm32` would be consistent with how the various CMake and Kconfig variables are defined (I suspect those get created based on `west.yml`).

This used to be an even bigger problem before: #73901, however even now I think it's important to be consistent and predictable between modules. We used to have the same bug in `hal_silabs`, but fixed it with https://github.com/zephyrproject-rtos/hal_silabs/commit/a09dd1b82b24aa3060e162c0dfa40026c0dba450

erwango:
@jhedberg I was just wanting to avoid impacting users for vain reason, but since you gave me valid ones, I'll do the change to `hal_stm32`.

jhedberg (approved):

PR #230, opened 2024-08-30
Introduce the STM32WB0 series

mathieuchopstm:
Introduce the STM32CubeWB0 package in hal_stm32 for the upcoming introduction of the STM32WB0 series support in Zephyr.

This PR includes two little hacks to work around a difference in the naming convention between STM32CubeWB0 and other Cube packages, that are clearly annotated with a **`HACK:`** comment:
* in `genllheaders.py`, replace `<stm32wb0xx_` with `<stm32wb0x_`) after rendering the common LL header to fix wrong filenames being used
* in `stm32cube/stm32wb0xx/CMakeLists.txt`, add `-DSTM32WB0n` define in addition to inappropriate `-DSTM32WB0nxx` from top-level `CMakeLists.txt`
  * (`n` here is the SoC number - `5`/`6`/`7`/`9`)
  * the `-DSTM32WB0yxx` define is not removed to avoid modifying top-level `CMakeLists.txt`
    * this should not bring any negative consequence


mathieuchopstm:
V2:
* remove `CMakeLists.txt` hack
  * Found not necessary after discussion
* rework `genllheaders.py` hack into a more generic patch
  * All series ending with a single `x` are now gracefully handled
* use the appropriate `stm32wb0x` name for the series, instead of `stm32wb0xx`

djiatsaf-st (approved):
LGTM

scripts/genpinctrl/stm32-pinctrl-config.yaml
```diff
@@ -42,6 +42,9 @@
 - name: ADC_IN / ADC_INN / ADC_INP
   match: "^ADC(?:\\d+)?_IN[NP]?\\d+$"
 
+- name: ADC_VINM / ADC_VINP
+  match: "^ADC(?:\\d+)?_VIN[PM]\\d+$"
```
erwango, line 46: Are we sure this won't have impact on other series ?
If this is the case, please split in another PR

stm32cube/stm32wb0x/README
```diff
@@ -0,0 +1,62 @@
...
+        . RAM_VR          (STM32WB0x virtual registers)
+    -in stm32cube/stm32wb0xx/soc/system_stm32wb0x.h
+      . Remove declaration of non-existent '__vector_table' symbol
+      . Remove declaration of associated datatypes 'intfunc' / 'intvec_elem'
+    -in stm32cube/stm32wb0xx/drivers/include/stm32wb0x_hal_conf.h
+      . Change USE_RADIO_PROPRIETARY_DRIVER from 1U to 0U
+         . Required for BLE stack compatibility
+    -in stm32cube/stm32wb0xx/drivers/src/stm32wb0x_hal_radio.c
+      . Remove definition of __blue_RAM variable
+        . Moved to 'zephyr/soc/st/stm32/stm32wb0xx/soc.c'
```
erwango, line 60: Please make at least one dedicated commit for this fixes so we can clearly identify them

erwango (changes requested):

scripts/genpinctrl/stm32-pinctrl-config.yaml
mathieuchopstm: I have done a blank `genpinctrl` run with latest STM32_open_pin_data and did not see `ADC_VINM / ADC_VINP` in any pin control file other than WB0, as expected, so I don't think it has any impact.

stm32cube/CMakeLists.txt
```diff
@@ -64,12 +65,22 @@ endif()
 foreach(stm_soc ${stm_socs})
   string(TOUPPER ${stm_soc} soc_to_upper)
   if(CONFIG_SOC_SERIES_${soc_to_upper})
+    # Check if we're building for a series ending with a single-'x', such as STM32WB0.
```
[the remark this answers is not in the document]
mathieuchopstm: OK, I will add a top-level comment to the CMakeLists to make this more clear.

mathieuchopstm:
v2:
* moved compatibility patches to separate commit
* added patch for `PAGESIZE` issue
* added comments to top-level CMakeLists.txt

stm32cube/stm32wb0x/drivers/src/stm32wb0x_hal_radio.c
```diff
@@ -204,15 +204,7 @@ static uint8_t CondRoutineRxTrue(ActionPacket *p)
...
 
-/* BLUE RAM, reserved for radio communication. Not usable from the application */
-__SECTION(".bss.__blue_RAM")
-#if defined(STM32WB05) || defined(STM32WB09)
-__REQUIRED(uint8_t __blue_RAM[CFG_NUM_RADIO_TASKS * 92 + 28]) = {0,};
-#elif defined(STM32WB06) || defined (STM32WB07)
-__REQUIRED(uint8_t __blue_RAM[CFG_NUM_RADIO_TASKS * 80 + 28]) = {0,};
-#else
-#warning "No Blue RAM allocated"
-#endif
```
erwango: For someone which may want to do further modification in those files, it maybe easier to have this defined under `#if 0` so that symbols are still there if one searches them.

erwango: Same for other patches

mathieuchopstm:
V3:
* reworked patches
  * `#if 0` blocks for deleted parts
  * comment for `CMakeLists.txt`
  * `USE_RADIO_PROPRIETARY_DRIVER` untouched
    * Patch is non-destructive and comment makes it self-explanatory

erwango (approved):

gautierg-st (approved):

PR #236, opened 2024-10-10
lib: stm32wb0: Add BLE controller implementation

HoZHel:
Provide BLE controller implementation for STM32WB0x SOCs.

Provide the required license (ST ULTIMATE LIBERTY) to use the binary BLOB.

Modify the yaml file to support fetching the STM32WB0 binary stack library
for Bluetooth implementation.


lib/CMakeLists.txt
```diff
@@ -70,4 +70,5 @@ if(CONFIG_HAS_STM32LIB)
     target_link_libraries(app PUBLIC stm32wba_ll_lib)
 
   endif()
+  add_subdirectory_ifdef(CONFIG_BT_STM32WB0 stm32wb0)
```
erwango, line 73: This is fine but we need consistency across vraious libs.
Hence I'd expect the same to be done fro WBA and WB

lib/stm32wb0/CMakeLists.txt
```diff
@@ -0,0 +1,309 @@
...
+zephyr_library_sources(BLE_TransparentMode/STM32_BLE/App/adv_buff_alloc.c)
+zephyr_library_sources(BLE_TransparentMode/System/Modules/RADIO_utils/Src/RADIO_utils.c)
+zephyr_library_sources(BLE_TransparentMode/System/Modules/miscutil.c)
+zephyr_library_sources(BLE_TransparentMode/STM32_BLE/App/pawr_buff_alloc.c)
+zephyr_library_sources(STM32_BLE/stack/config/ble_stack_user_cfg.c)
+zephyr_library_sources(BLE_TransparentMode/STM32_BLE/App/dtm_cmds.c)
+zephyr_library_sources(BLE_TransparentMode/Core/Src/stm32wb0x_hal_msp.c)
+
+else()
+zephyr_compile_definitions(-DCFG_BLE_NUM_RADIO_TASKS=0)
```
erwango, line 308: Do you expect cases where `CONFIG_BT_STM32WB0=y` and `CONFIG_BT=n` ?

erwango (changes requested):
A README file summing up the origin of the files added would be required.
Similar to https://github.com/zephyrproject-rtos/hal_stm32/blob/main/lib/stm32wb/hci/README.

Also, to simplify follow up maintenance, I'd expect these files could be updated automatically using [existing script](https://github.com/zephyrproject-rtos/hal_stm32/blob/main/scripts/ble_library.py). This is not a blocker, but it would be required to have it eventually.

HoZHel:
> A README file summing up the origin of the files added would be required. Similar to https://github.com/zephyrproject-rtos/hal_stm32/blob/main/lib/stm32wb/hci/README.

Ok, I'll add it.

> Also, to simplify follow up maintenance, I'd expect these files could be updated automatically using [existing script](https://github.com/zephyrproject-rtos/hal_stm32/blob/main/scripts/ble_library.py). This is not a blocker, but it would be required to have it eventually.

I will update this file in the future.

lib/CMakeLists.txt
HoZHel: I didn't get this comment well. Should I do something?

lib/stm32wb0/CMakeLists.txt
HoZHel: No, you're right. I'll fix it.

lib/CMakeLists.txt
erwango: You're adding a sub section to handle WB0 case, which is indeed a good idea.
But in that case, for consistency, it would be nice to do the same on other series (in a dedicated commit).

lib/stm32wb0/BLE_TransparentMode/Core/Src/stm32wb0x_hal_msp.c
```diff
@@ -0,0 +1,369 @@
...
+  /* USER CODE END USART1_MspInit 0 */
+    /* Peripheral clock enable */
+    __HAL_RCC_USART1_CLK_ENABLE();
+
+    __HAL_RCC_GPIOB_CLK_ENABLE();
+    __HAL_RCC_GPIOA_CLK_ENABLE();
+    /**USART1 GPIO Configuration
+    PB0     ------> USART1_RX
+    PA1     ------> USART1_TX
+    */
```
erwango, line 233: Is that actually required ?
Don't you rely on Zephyr to enable USART1 ?

lib/stm32wb0/BLE_TransparentMode/Core/Src/main.c
erwango, line 1: Do you actually require this file ?

I don't see it being compiled ?
In general, only include files that you strictly need

lib/stm32wb0/README
```diff
@@ -0,0 +1,123 @@
...
+      Projects/NUCLEO-WB09KE/Applications/BLE/BLE_TransparentMode/STM32_BLE/App/adv_buff_alloc_tiny.c
+      Projects/NUCLEO-WB09KE/Applications/BLE/BLE_TransparentMode/STM32_BLE/App/transport_layer.c
+      Projects/NUCLEO-WB09KE/Applications/BLE/BLE_TransparentMode/STM32_BLE/App/dtm_preprocess_events.c
+      Projects/NUCLEO-WB09KE/Applications/BLE/BLE_TransparentMode/STM32_BLE/App/adv_buff_alloc.c
+      Projects/NUCLEO-WB09KE/Applications/BLE/BLE_TransparentMode/System/Modules/RADIO_utils/Src/RADIO_utils.c
+      Projects/NUCLEO-WB09KE/Applications/BLE/BLE_TransparentMode/System/Modules/miscutil.c
+      Projects/NUCLEO-WB09KE/Applications/BLE/BLE_TransparentMode/STM32_BLE/App/pawr_buff_alloc.c
+      Projects/NUCLEO-WB09KE/Applications/BLE/BLE_TransparentMode/STM32_BLE/App/dtm_cmds.c
+      Projects/NUCLEO-WB09KE/Applications/BLE/BLE_TransparentMode/Core/Src/stm32wb0x_hal_msp.c
+      Middlewares/ST/STM32_BLE/stack/config/ble_stack_user_cfg.c
```
erwango, line 77: I see more than those files in the commit (Projects/NUCLEO-WB09KE/Applications/BLE/BLE_TransparentMode/Core  files for instance).
Can you verify ?

Also it would be easier if the list was sorted.

erwango (changes requested):
Thanks for the CMake clean up.
So more comments on the files extracted from the lib to here. Are they all required ?

lib/stm32wb0/README
HoZHel: Actually, it was @msmttchr's idea to deliver the library with as less changes as possible so as to make it easier for future deliveries.

Ok, I'll sort them.

erwango: > Actually, it was @msmttchr's idea to deliver the library with as less changes as possible so as to make it easier for future deliveries.

Minimal changes, I fully agree.
But, till now, for WB and WBA we've only pushed the minimal set of required files to ease the update and review process. Note that using the script this isn't too much of a hassle.

lib/stm32wb0/README
```diff
@@ -0,0 +1,123 @@
...
+   https://github.com/STMicroelectronics/STM32CubeWB0
+
+Commit:
+   61d80e00cdb6136a58a33b95d2137e3bc9baa28e
+
+Maintained-by:
+   Internal
+
+License:
+   ULTIMATE LIBERTY SOFTWARE LICENSE AGREEMENT
```
erwango, line 93: License of files included this module should be OSI approved (BSD-3 for instance)

erwango (changes requested):

lib/stm32wb0/LICENSE.md
```diff
@@ -0,0 +1,27 @@
+Copyright 2019-2021 STMicroelectronics.
```
erwango, line 1:
```suggestion
Copyright XXXXXX STMicroelectronics.
```

lib/stm32wb0/README
HoZHel: Fixed.

lib/stm32wb0/LICENSE.md
HoZHel: Done.

lib/stm32wb0/README
HoZHel: Done.

lib/stm32wb0/README
```diff
@@ -0,0 +1,127 @@
...
+   It is available in stm32cube/stm32wb0x/drivers
+
+URL:
+   https://github.com/STMicroelectronics/STM32CubeWB0
+
+Commit:
+   61d80e00cdb6136a58a33b95d2137e3bc9baa28e
+
+Maintained-by:
+   Internal
```
erwango, line 92: This is from Zephyr perspective. 
`Internal` would be zephyr project, so it should be `External`

erwango (changes requested):
Thanks for the changes.
One minor change and we're good to go.

lib/stm32wb0/README
HoZHel: Done.

lib/stm32wb0/BLE_TransparentMode/Core/Src/stm32wb0x_hal_msp.c
HoZHel: `HAL_UART_MspInit` and `HAL_UART_MspDeInit `have been excluded from compilation. To minimize the changes I used `#ifndef __ZEPHYR__`

lib/stm32wb0/BLE_TransparentMode/Core/Src/main.c
HoZHel: Redundant files have been removed.

erwango (approved):

HoZHel:
Rebased
Removed trailing white spaces
dos2unix applied

PR #241, opened 2024-12-01
scripts: genpinctrl: add TSC pinctrl definition

arifbalik:
Added TSC regex to match and generate pin definitions. 

I'm not sure if I should also put here the new pinctrl files, but I've run the script and it generates the pins

### DIFF
--- a/scripts/genpinctrl/stm32-pinctrl-config.yaml
+++ b/scripts/genpinctrl/stm32-pinctrl-config.yaml
@@ -299,3 +299,6 @@
 
 - name: JTAG PORT
   match: ^(SYS|DEBUG)_((JTMS-)?SWDIO|(JTCK-)?SWCLK|JTDI|JTDO(-TRACESWO|-SWO)?|(NJ)?JTRST)$
+
+- name: TSC
+  match: "^TSC_(?:G\\d+_IO\\d+|SYNC)$"
--- a/scripts/genpinctrl/stm32-pinctrl-config.yaml
+++ b/scripts/genpinctrl/stm32-pinctrl-config.yaml
@@ -259,6 +259,9 @@
 - name: TIM_CH / TIM_CHN
   match: "^TIM\\d+_CH\\d+N?$"
 
+- name: TSC
+  match: "^TSC_(?:G\\d+_IO\\d+|SYNC)$"
+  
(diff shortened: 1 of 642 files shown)


scripts/genpinctrl/stm32-pinctrl-config.yaml
```diff
@@ -299,3 +299,6 @@
 
 - name: JTAG PORT
   match: ^(SYS|DEBUG)_((JTMS-)?SWDIO|(JTCK-)?SWCLK|JTDI|JTDO(-TRACESWO|-SWO)?|(NJ)?JTRST)$
+
+- name: TSC
+  match: "^TSC_(?:G\\d+_IO\\d+|SYNC)$"
```
erwango, line 304: Please sort alphabetically. Seems that the alphabetical sorting requires some review, but let's not making it worse.

erwango (commented):
Thanks. Looks fine, but please split template changes from the new batch of generated files in a dedicated commit.

scripts/genpinctrl/stm32-pinctrl-config.yaml
arifbalik: Do you mean to move this definition somewhere in the file where `name` is in alphabetical order? or the generated pin definitions?

erwango: > Do you mean to move this definition somewhere in the file where name is in alphabetical order? 

yup

arifbalik: Got it, thanks

arifbalik:
> @arifbalik I think @erwango meant split it into different commits, not into different PRs, anyway same result if we merge them in sequence, @erwango how do you want this?

Ah, my bad, I can put the commit here as well if you like

fabiobaltieri:
> Ah, my bad, I can put the commit here as well if you like

If you don't mind, better to keep them under the same PR I think

fabiobaltieri (approved):

erwango (approved):
Thanks, I meant same PR but different commits indeed.
Next time, please put commits in logical order (template changes first, then new batch of files). Anyway not blocking, so "approved", thanks.

PR #244, opened 2024-12-04
Align lib/stm32wba with  STM32CubeWBA structure and with BLE_TransparentMode application

benothmn-st:
Align the lib/stm32wba folder structure with STM32CubeWBA structure
Align the code with BLE_TransparentMode application from STM32CubeWBA
Align the scripts/ble_library.py with the new structure of lib/stm32wba


lib/stm32wba/BLE_TransparentMode/Core/Inc/app_conf.h
```diff
@@ -130,7 +65,7 @@
 #define CFG_BLE_OPTIONS             (0 | \
                                      0 | \
                                      0 | \
-              BLE_OPTIONS_EXTENDED_ADV | \
+                                     0 | \
```
asm5878, line 68: BLE_OPTIONS_EXTENDED_ADV should be added to have same BLE features enabled

lib/stm32wba/BLE_TransparentMode/Core/Inc/app_conf.h
```diff
@@ -444,15 +372,15 @@ typedef enum
 #define NEXT_EVENT_SCHEDULING_FROM_ISR      (1)
 
 /* Link Layer uses temperature based calibration (0 --> NO ; 1 --> YES) */
-#define USE_TEMPERATURE_BASED_RADIO_CALIBRATION  (0)
+#define USE_TEMPERATURE_BASED_RADIO_CALIBRATION  (1)
```
asm5878, line 375: Calibration must be reworked according latest Cube, better to disable it

lib/stm32wba/BLE_TransparentMode/Core/Inc/app_conf.h
```diff
@@ -444,15 +372,15 @@ typedef enum
...
 
 #define RADIO_INTR_NUM                      RADIO_IRQn     /* 2.4GHz RADIO global interrupt */
 #define RADIO_INTR_PRIO_HIGH                (0)            /* 2.4GHz RADIO interrupt priority when radio is Active */
 #define RADIO_INTR_PRIO_LOW                 (5)            /* 2.4GHz RADIO interrupt priority when radio is Not Active - Sleep Timer Only */
 
 #if (USE_RADIO_LOW_ISR == 1)
 #define RADIO_SW_LOW_INTR_NUM               HASH_IRQn      /* Selected interrupt vector for 2.4GHz RADIO low ISR */
-#define RADIO_SW_LOW_INTR_PRIO              (14)           /* 2.4GHz RADIO low ISR priority */
+#define RADIO_SW_LOW_INTR_PRIO              (15)           /* 2.4GHz RADIO low ISR priority */
```
asm5878, line 383: I remember 14 was required to avoid an assert failure in a priority check. Have you tried to run an application to verify the behavior?

asm5878 (changes requested):
We should align stm32\scripts\ble_library.py to be aligned with the new structure. Do you agree?

lib/stm32wba/BLE_TransparentMode/Core/Inc/app_conf.h
benothmn-st: ok, I will add it, thanks!

lib/stm32wba/BLE_TransparentMode/Core/Inc/app_conf.h
benothmn-st: As you said, it's better for keeping the same behavior to keep it 14 .
We need to review all the app_conf file and put some of them as build parameters.

benothmn-st:
> We should align stm32\scripts\ble_library.py to be aligned with the new structure. Do you agree?

I thought to add it in a separate PR to go faster, but I am okay with your proposal also. I will do it.

erwango:
Sorry for the noise with Copilot. Please discard

asm5878 (approved):
All seems fine at my side

lib/stm32wba/BLE_TransparentMode/Core/Inc/app_conf.h
```diff
@@ -419,7 +347,7 @@ typedef enum
  *   - 2 : Debugger available in low power mode.
  *
  ******************************************************************************/
-#define CFG_DEBUGGER_LEVEL           (0)
```
erwango, line 422: Is that expected to enable debugger by default ?
Is there an impact to be expected on power consumption?

scripts/ble_library.py
```diff
@@ -193,16 +193,20 @@ def copy_hci_files(src_repo_path, dest_lib_path, stm32_serie):
...
         make_version: latest STM32Cube version.
         make_commit: Commit corresponding to latest STM32Cube version.
+        stm32_serie: stm32 serie
     """
 
-    readme_path = Path(lib_path / "hci" / "README")
+    if stm32_serie == "stm32wba":
+        readme_path = Path(lib_path / "README")
+    else:
+        readme_path = Path(lib_path / "hci" / "README")
```
erwango, line 209: Note: For this last change, I'd actually prefer we move the WB README file to align it on WBA/WB0

erwango (changes requested):

lib/stm32wba/BLE_TransparentMode/Core/Inc/app_conf.h
asm5878: The goal of this PR is trying to have same code/features but with a new structure. 
And in this case we need to leave the same debugger conf (it means "0").
We could make same comments for other features (LED, BUTTONS, PM,...), but after this PR will be merged a new PR will be done to review systematically all the aspects of the integration

erwango: So, it meas that users will not enjoy the same behavior once this PR is merged and until another one is done. Not sure it is acceptable

erwango: This change should not introduce functional impact.

asm5878: I was no  clear in my argumentations, my mistake. With this PR we are not introducing any functional changement. Just rearranging the code according a proper directory structure using the Cube BLE Transparent mode application as a reference. As already done for STM32WB0. Enhancements will be introduced when available. This code rearrangement is mandatory to better highlight Cube FW code we are reusing in Zephyr. Hoping now it's enough clear.

asm5878: Just adding some configuration defines are inherited from Cube  (as CFG_LED_SUPPORTED or CFG_DEBUGGER_LEVEL) but are not used at all inside Zephyr integration. So their values have no impact at all on behavior.

erwango (approved):

asm5878 (approved):
Fine for me

PR #245, opened 2024-12-04
Update  pincrtl.dsti files for all stm32 family 

djiatsaf-st:
- introduce new *-pincrtl.dtsi files for c051, c071,
c091, c092, wba5, g411, g414, g431, g491 and g473 series.
- delete pins or add new pins in others series


erwango (dismissed):
Please update dts/README.rst

FRASTM (approved):

dts/st/f3/stm32f301c(6-8)tx-pinctrl.dtsi
```diff
@@ -686,10 +686,22 @@
 				pinmux = <STM32_PINMUX('A', 0, AF3)>;
 			};
 
+			/omit-if-no-ref/ tsc_g1_io1_pa0: tsc_g1_io1_pa0 {
+				pinmux = <STM32_PINMUX('A', 0, AF3)>;
+			};
+
```
gautierg-st, line 692: All these tsc pins are duplicated. Any ideas why? Does it come from the XML source or is it a problem with our script?

gautierg-st (changes requested):
There is a problem with the TSC pins generation. They are duplicated for many series (F0, F3, L0, L4, L5, U0, U5, WB, WB0, WBA).

dts/st/f3/stm32f301c(6-8)tx-pinctrl.dtsi
erwango: It seems that TSC filer is present twice.... This should be fixed

gautierg-st (approved):

erwango (approved):

PR #247, opened 2024-12-13
Add SWD pinmux to stm32f1 family

alteman:
Add missing SWD pinmux definitions for STM32F1 MCU family

### DIFF
--- a/scripts/genpinctrl/stm32f1-pinctrl-config.yaml
+++ b/scripts/genpinctrl/stm32f1-pinctrl-config.yaml
@@ -170,6 +170,10 @@
   match: "^I2S\\d+_SD$"
   mode: alternate
 
+- name: JTAG PORT
+  match: "^(SYS|DEBUG)_((JTMS-)?SWDIO|(JTCK-)?SWCLK|JTDI|JTDO(-TRACESWO|-SWO)?|(NJ)?JTRST)$"
+  mode: alternate
+
 - name: RCC_MCO
   match: "^RCC_MCO_?(\\d+)?$"
   mode: alternate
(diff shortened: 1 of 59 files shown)


scripts/genpinctrl/stm32f1-pinctrl-config.yaml
```diff
@@ -256,6 +256,10 @@
   match: "^USB_OTG_FS_(?:DM)?(?:DP)?(?:SOF)?(?:ID)?(?:VBUS)?$"
   mode: input
 
+- name: JTAG PORT
+  match: ^(SYS|DEBUG)_((JTMS-)?SWDIO|(JTCK-)?SWCLK|JTDI|JTDO(-TRACESWO|-SWO)?|(NJ)?JTRST)$
+  mode: alternate
+
```
gautierg-st, line 262: Nit: Move it just after I2S to keep alphabetical order

gautierg-st (changes requested):
Could you also regenerate the pinctrl.dtsi so that the SWD pins are included in them (that also verifies that the regex is good).

scripts/genpinctrl/stm32f1-pinctrl-config.yaml
erwango: This is done on top of the branch. You just need to rebase on top of https://github.com/zephyrproject-rtos/hal_stm32/commit/37842371f5ef0078ad32f16e5059c1df58b51892

alteman:
- Moved the section to follow I2S
- Regenerated the .dtsi files as a separate commit
- Rebased onto current main/HEAD (37842371)

On an unrelated note: mcu/STM32WBA50KEUx.xml has been deleted from STM32_open_pin_data in commit f4ec11f00e762e37ffc4020f6d4f20d225bc061d (STM32CubeMX 6.12.0), but dts/st/wba/stm32wba50keux-pinctrl.dtsi is still alive in this repo. Not sure if that's by design.

gautierg-st (approved):

erwango (approved):

PR #248, opened 2024-12-18
stm32cube: Modify SAI_TypeDef U585

chrisschir:
Changed SAI_TypeDef of U585 to consist
of two SAI_Block_TypeDef

Signed-off-by: Christopher Schirk <christopher.schirk@draeger.com>

### DIFF
--- a/stm32cube/stm32u5xx/soc/stm32u585xx.h
+++ b/stm32cube/stm32u5xx/soc/stm32u585xx.h
@@ -1213,23 +1213,24 @@ typedef struct
   */
 typedef struct
 {
-  __IO uint32_t GCR;          /*!< SAI global configuration register,        Address offset: 0x00 */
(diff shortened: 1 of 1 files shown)


erwango:
@chrisschir Thanks for this contribution, but this repo aims at being a "fork" of STM32Cube packages so we don't modify HAL if we know modification won't be integrated in Cube package. Though we can merge fixes waiting for official Cube package delivery.

The change you proposed doesn't seem to fall in the bug fix category and the impact of the change may be larger than intended.
Would you min explaining what was the issue that lead you to this change and why it can only be fixed this way?

In any case, I'd suggest to open the issue in https://github.com/STMicroelectronics/STM32CubeU5 and explain why it is required.

erwango:
Closing as per https://github.com/STMicroelectronics/STM32CubeU5/issues/56

PR #253, opened 2025-01-21
lib: stm32wba: Clean unused code which was implemented in zephyr repo

benothmn-st:
Remove all the unused source code that was implemented in zephyr repo

### DIFF
--- a/lib/stm32wba/BLE_TransparentMode/STM32_WPAN/Target/linklayer_plat.c
+++ b/lib/stm32wba/BLE_TransparentMode/STM32_WPAN/Target/linklayer_plat.c
@@ -4,38 +4,12 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-
-
-/* USER CODE END Header */
-
-#ifndef __ZEPHYR__
-#include "stm32wbaxx_hal.h"
-#include "stm32wbaxx_hal_conf.h"
-#include "stm32wbaxx_ll_rcc.h"
-
-#include "app_common.h"
-#include "app_conf.h"
-#include "linklayer_plat.h"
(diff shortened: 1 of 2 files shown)


erwango (changes requested):
I'm fine with clean up, but I have 2 questions around it
- how do we ensure that this code won't be overwritten at next Cube update ?
- We have the same unused code on WB0, do you plan to make similar clean up ? Or is it a different use case

In any case, please update README, so that we know the diif to be maintained at next cube update if update script fails

benothmn-st:
> I'm fine with clean up, but I have 2 questions around it
> 
> * how do we ensure that this code won't be overwritten at next Cube update ?
> * We have the same unused code on WB0, do you plan to make similar clean up ? Or is it a different use case
> 
> In any case, please update README, so that we know the diif to be maintained at next cube update if update script fails

Hi Erwan,

With Alessandro, we agree to keep the code as it is actually. So, the comparison with the Cube FW still easy to do.
We will add in the next PR of Cube FW update a comment (in the first #ifndef __ZEPHYR__) to mention that all those functions that are not compiled here are redefined in the zephyr repo and we will mention the file as reference.
If you agree with this proposition, you can close the PR.

Thanks!

erwango:
> With Alessandro, we agree to keep the code as it is actually. So, the comparison with the Cube FW still easy to do.
We will add in the next PR of Cube FW update a comment (in the first #ifndef ZEPHYR) to mention that all those functions that are not compiled here are redefined in the zephyr repo and we will mention the file as reference.
If you agree with this proposition, you can close the PR.

I'm fine with this proposal. Thanks
I'm closing the PR then.

PR #254, opened 2025-01-30
SAI pinctrl

yishai1999:
Added SAI pinctrl nodes
I am planning on writing an SAI I2S driver in zephyr so I need these nodes to be defined.
I'm not sure how to see if I should add more properties (pull-up, push-pull...). So LMK if I'm making some mistake.

BTW, I'm getting the following warnings when running the genpinctrl script:
```
Unexpected AF format: GPIO_AF2_rf (ip: STM32WL33_gpio_v1_0)                                                                                                                                                 
Skipping unsupported family STM32MP2.                                                                                                                                                                       
Skipping unsupported family STM32U3.                                                                  
Skipping unsupported family STM32WL3.
```

### DIFF
--- a/dts/st/f4/stm32f413c(g-h)ux-pinctrl.dtsi
+++ b/dts/st/f4/stm32f413c(g-h)ux-pinctrl.dtsi
@@ -614,6 +614,28 @@
 				slew-rate = "very-high-speed";
 			};
 
+			/* SAI */
+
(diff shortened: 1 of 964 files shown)


gautierg-st (approved):
Regex for pin generation looks good to me.
I'm not very familiar with SAI to comment on pin properties but in example project in STM32CubeIDE, there are no pull-ups, or open-drain and speed is low. So I don't think it is necessary to add anything here.

erwango (approved):

PR #256, opened 2025-01-30
Stm32n6xx: Uncomment the list of HAL driver modules

marwaiehm-st:
Uncommented the list of HAL driver modules to ensure consistency with other STM32 series.


gautierg-st (approved):

erwango (approved):
Thanks

PR #258, opened 2025-02-05
Ethernet: Add stm32n6 pinctrl

marwaiehm-st:
Update STM32N6 pinctrl dtsi files with the ETH pins
Add STM32N6 Ethernet pinctrl config to stm32-pinctrl-config.yaml


scripts/genpinctrl/stm32-pinctrl-config.yaml
```diff
@@ -144,6 +144,22 @@
   match: "^ETH_TXD3$"
   slew-rate: very-high-speed
 
+- name: ETH
```
erwango, line 147: This might not be obvious why this differs from other ETH filters above.
Maybe you should specify "ETH (STM32N6 IP)", or sort of.

Additionally, it would be nice to squash the other ETH pins into a one liner to help comparing (non-blocking though)

erwango (changes requested):
Please put commits in the logical order:
- New filters
- New batch

scripts/genpinctrl/stm32-pinctrl-config.yaml
marwaiehm-st: Done

erwango (approved):

gautierg-st (approved):

PR #264, opened 2025-03-04
scripts: stm32cube: genpinctrl for  debug_njtrst pinout

FRASTM:
Generates the debug_njtrst_pb4 pin for the stm32wba65x serie where the PB4(NJRST) is named "DEBUG_NJTRST"

from **STM32WBA65RIVx.xml**
```
    <Pin Name="PB4(NJRST)" Position="32" Type="I/O">
        <Signal Name="DEBUG_NJTRST"/>
```

to **stm32wba65rivx-pinctrl.dtsi**
```
			/omit-if-no-ref/ debug_njtrst_pb4: debug_njtrst_pb4 {
				pinmux = <STM32_PINMUX('B', 4, AF0)>;
			};
```

PR #265, opened 2025-03-04
stm32 dts new pinctrl description files based on the new  open_pin_data STM32CubeMX 6.14.0

FRASTM:
Generates the pinctrl .dtsi from the https://github.com/STMicroelectronics/STM32_open_pin_data  6.14.0

--> Will also define **debug_njtrst_pb4** that where previously missing

			/omit-if-no-ref/ debug_njtrst_pb4: debug_njtrst_pb4 {
				pinmux = <STM32_PINMUX('B', 4, AF0)>;
			};

thanks to the https://github.com/zephyrproject-rtos/hal_stm32/pull/264


FRASTM:
see internal ticket https://intbugzilla.st.com/show_bug.cgi?id=204691

gmarull (approved):

asm5878:
@FRASTM I guess stm32wba5x dtsi files have been not properly updated to manage pin renaming from debug_jtrst_pb4 to  debug_njtrst_pb4. Am I wrong ?

PR #275, opened 2025-04-18
scripts: genpinctrl: add support for LPTIM pins.

idekine:
Add support for LPTIM pins.


erwango:
Can you provide some context on this PR ?
So far, I don't think we have anything that could use this in zephyr/ public repo. Do you have further plans around this ?

PR #276, opened 2025-04-24
Adds HAL support for STM32F401xD variants

rriveramcrus:
Modifies the stm32f4xx HAL files to add support for the STM32F401xD family of devices. Since STM32F401xE family shares a datasheet with the STM32F401xD family, they are assumed to be similar enough to share the HAL code.

Special attention should be given to the changes made to `stm32cube/stm32f4xx/drivers/include/stm32f4xx_hal_flash_ex.h` since this is where the D and E families may differ, but I expect the flash IP to be mostly the same.


rriveramcrus:
https://github.com/STMicroelectronics/STM32CubeF4/issues/192
closing this based on the discussion here

erwango:
@rriveramcrus We'll need to fix following line https://github.com/zephyrproject-rtos/zephyr/blob/b450c71e4cc083f19880fd36b0ec554b0a756ee0/soc/st/stm32/stm32f4x/Kconfig.soc#L107. Would you mind pushing this change ?

PR #279, opened 2025-05-07
dts: -pinctrl.dtsi: Label clean up and HSPI signals

erwango:
- Rename ETH N6 IP label to be less confusing when generated on other series.
- Generate ETH extended signals on STM32MP13
- Generate HSPI signals (STM32U5)

### DIFF
--- a/scripts/genpinctrl/stm32-pinctrl-config.yaml
+++ b/scripts/genpinctrl/stm32-pinctrl-config.yaml
@@ -111,6 +111,10 @@
 - name: HRTIM_SCIN / HRTIM_SCOUT
   match: "^HRTIM\\d+_SC(IN|OUT)$"
 
+- name: HSPI
+  match: "^HSPI(.*)(?:CLK|NCS|DQS|IO\\d+)$"
+  slew-rate: very-high-speed
+
 - name: I2C_SCL
   match: "^I2C\\d+_SCL$"
   drive: open-drain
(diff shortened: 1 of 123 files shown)


etienne-lms (commented):
For commit "dts: New batch with new ETH label and MP13 signals":
Would be nice to be plit in 2 commits: one for the ETH label and one for adding MP13 pinctrl nodes.

For commit "dts: New batch with HSPI signals":
There are more than only HSPI related changes in this commit. Could you move the JTAG port related changes into a specific commit?

scripts/genpinctrl/stm32-pinctrl-config.yaml
```diff
@@ -247,4 +247,3 @@
 
 - name: USB
   match: "^USB_(?:DM)?(?:DP)?(?:NOE)?$"
-
```
etienne-lms, line 250: Nitpicking: I think you should discard the change on this last file line.

etienne-lms (commented):
Comment for commits "dts: New batch with new ETH label" and "dts: ETH MP13 signals":
The 2 commit messages only mention ETH but also add data related to SAI and a few UART related devices. I would propose you squash the 2 commits and update the commit messages to mention also updates of SAI and UART related devices.

scripts/genpinctrl/stm32-pinctrl-config.yaml
erwango: It should not have been added and was removed by my editor. I'll keep it that way.

etienne-lms (approved):

gautierg-st (approved):

PR #281, opened 2025-05-18
hal_stm32: fix SDIO initialization failure when skipping card identification sequence.

ExaltZephyr:
This PR fixes an issue in SDIO driver where HAL_SDIO_Init fails if the card identification sequence is skipped.


erwango (changes requested):
Please document changes in each HAL README file, otherwise change is ok and will be delivered in HAL in next versions.

PR #283, opened 2025-06-06
lib/stm32: stm32wba: radio temperature calibration activation

rjayles:
This work on the lib/stm32 allows to activate/deactivate the calibration of the radio based on the temperature sensor.
It can be activated/deactivated by a Zephyr config

### DIFF
--- a/lib/stm32wba/BLE_TransparentMode/Core/Inc/app_conf.h
+++ b/lib/stm32wba/BLE_TransparentMode/Core/Inc/app_conf.h
@@ -376,7 +376,11 @@ typedef enum
  * HW RADIO configuration
  ******************************************************************************/
 /* Link Layer uses temperature based calibration (0 --> NO ; 1 --> YES) */
-#define USE_TEMPERATURE_BASED_RADIO_CALIBRATION  (0)
+#ifdef CONFIG_BT_STM32WBA_USE_TEMPERATURE_BASED_RADIO_CALIBRATION
+  #define USE_TEMPERATURE_BASED_RADIO_CALIBRATION  (1)
+#else
+  #define USE_TEMPERATURE_BASED_RADIO_CALIBRATION  (0)
+#endif /*CONFIG_BT_STM32_USE_TEMPERATURE_BASED_RADIO_CALIBRATION*/
 
 #define RADIO_INTR_NUM                      RADIO_IRQn     /* 2.4GHz RADIO global interrupt */
 #define RADIO_INTR_PRIO_HIGH                (0)            /* 2.4GHz RADIO interrupt priority when radio is Active */
--- a/lib/stm32wba/BLE_TransparentMode/STM32_WPAN/Target/linklayer_plat.c
+++ b/lib/stm32wba/BLE_TransparentMode/STM32_WPAN/Target/linklayer_plat.c
@@ -28,10 +28,10 @@
 #include "linklayer_plat.h"
 #include "scm.h"
 #include "log_module.h"
+#ifndef __ZEPHYR__
 #if (USE_TEMPERATURE_BASED_RADIO_CALIBRATION == 1)
 #include "adc_ctrl.h"
 #endif /* (USE_TEMPERATURE_BASED_RADIO_CALIBRATION == 1) */
-#ifndef __ZEPHYR__
 #if (CFG_LPM_LEVEL != 0)
 #include "stm32_lpm.h"
 #include "stm32_lpm_if.h"
@@ -502,6 +502,7 @@ void LINKLAYER_PLAT_RCOStopClbr(void)
   while (LL_PWR_IsActiveFlag_VOS() == 0);
 #endif /* (CFG_SCM_SUPPORTED == 1) */
 }
+#endif /*__ZEPHYR__*/
 
 /**
   * @brief  Link Layer requests temperature.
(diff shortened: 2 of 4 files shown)


lib/stm32wba/BLE_TransparentMode/Core/Inc/app_conf.h
```diff
@@ -376,7 +376,11 @@ typedef enum
  * HW RADIO configuration
  ******************************************************************************/
 /* Link Layer uses temperature based calibration (0 --> NO ; 1 --> YES) */
-#define USE_TEMPERATURE_BASED_RADIO_CALIBRATION  (0)
+#ifdef CONFIG_BT_STM32WBA_USE_TEMPERATURE_BASED_RADIO_CALIBRATION
+  #define USE_TEMPERATURE_BASED_RADIO_CALIBRATION  (1)
+#else
+  #define USE_TEMPERATURE_BASED_RADIO_CALIBRATION  (0)
+#endif /*CONFIG_BT_STM32_USE_TEMPERATURE_BASED_RADIO_CALIBRATION*/
```
erwango: I see 2 issues here:
- CONFIG_BT_STM32WBA_USE_TEMPERATURE_BASED_RADIO_CALIBRATION is not defined 
- All code imported from Cube should not depend from Zephyr.

One suggestion would be to modify Cube code this way:
```
#ifndef USE_TEMPERATURE_BASED_RADIO_CALIBRATION
#define USE_TEMPERATURE_BASED_RADIO_CALIBRATION (0)
#endif
```

Then, define USE_TEMPERATURE_BASED_RADIO_CALIBRATION the way it is done for `HSE_VALUE` below:
https://github.com/zephyrproject-rtos/hal_stm32/blob/d49fb6dbe780b931eee33cce8656e87c46070fa8/stm32cube/CMakeLists.txt#L49-L51

Ideally, this way, change could be integrated in Cube library so we don't have to maintain this as a Zephyr fix.

Finally, don't forget to document the fix in related README file.

erwango (changes requested):

lib/stm32wba/BLE_TransparentMode/Core/Inc/app_conf.h
rjayles: Thanks for your comment, this way of working has been integrated and pushed

asm5878 (approved):

lib/stm32wba/CMakeLists.txt
```diff
@@ -60,6 +60,11 @@ set_target_properties(
   stm32wba_ble_lib PROPERTIES IMPORTED_LOCATION ${STM32WBA_BLE_LIB_DIR}/stm32wba_ble_stack_llo.a
 )
 
+# Setting the right Cube define according to the Zephyr configuration
+if(NOT CONFIG_BT_STM32WBA_USE_TEMPERATURE_BASED_RADIO_CALIBRATION STREQUAL "y")
```
erwango: Please add a Zephyr PR to add the definition of this new symbol and points to this PR (through west.yml update)

lib/stm32wba/README.rst
```diff
@@ -109,7 +109,7 @@ Patch List:
 	* Enabled extended advertising in CFG_BLE_OPTIONS:
 	  Impacted file: app_conf.h
 
-	* Disable Temperature based radio calibration:
+	* Allowing Temperature based radio calibration configuration to be externally overwritten:
 	  Impacted file: app_conf.h
```
erwango, line 116: Can you confirm there is a plan to integrate this in Cube HAL ?
@msmttchr FYI

lib/stm32wba/CMakeLists.txt
rjayles: The Zephyr PR has been created : https://github.com/zephyrproject-rtos/zephyr/pull/91898

lib/stm32wba/README.rst
rjayles: This topic has been discussed, and Cube modification will be done to take into account the Zephyr integration

benothmn-st (approved):

lib/stm32wba/README.rst
erwango, line 116: I see more files impacted than single app_conf.h. Can you update this list ?

erwango (changes requested):

lib/stm32wba/README.rst
rjayles: The modification in ll_sys_if.c and ll_sys.h is now covered. The modification in linklayer_plat.c is already covered with the following patch "#ifndef __ZEPHYR__ applied to remove Cube specific implementation Impacted files"

lib/stm32wba/CMakeLists.txt
```diff
@@ -60,6 +60,11 @@ set_target_properties(
...
+  zephyr_compile_definitions( -DUSE_TEMPERATURE_BASED_RADIO_CALIBRATION=0 )
+endif()
```
mathieuchopstm: Nit: IMO, the default should be 0 and we should set =1 only if Kconfig is enabled.

Also - **but I may be wrong on that** - I think this should be simply:
`if (NOT CONFIG_BT_STM32WBA_USE_TEMPERATURE_BASED_RADIO_CALIBRATION)`

mathieuchopstm (approved):

erwango (approved):

PR #286, opened 2025-06-13
STM32MP2: fix the USE_HAL_DRIVER include

youssefz24:
Modify the included HAL header in stm32mp2xx.h to the correct one. "stm32mp2xx_hal_conf.h" -> "stm32mp2xx_hal.h"

"stm32mp2xx_hal.h" was never included in build time and caused declaration errors when using the HAL driver.

This was discovered when trying to update the hwinfo driver to pass CI_tests for [STM32MP2 integration in Zephyr](https://github.com/zephyrproject-rtos/zephyr/pull/90295#) which requires hal functions.

### DIFF
--- a/stm32cube/stm32mp2xx/README
+++ b/stm32cube/stm32mp2xx/README
@@ -53,4 +53,10 @@ Patch List:
 	 assertion checks are not implemented, as the code still functions
 	 correctly without them.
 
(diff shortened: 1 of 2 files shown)


arnopo (approved):
Coherent with other MCU/MPU stm32 HAL

PR #287, opened 2025-06-14
fix: scripts: mp1 series eth pin config

arifbalik:
STM32MP1 series has different Ethernet namings than what is specified in stm32-pinctrl-config.yaml, so I did put another config file with one regex; `ETH[12]_(CLK|COL|CRS|CRS_DV|GTX_CLK|MDC|MDIO|PHY_INTN|PPS_OUT|REF_CLK|RX_CLK|RX_CTL|RX_DV|RX_ER|RXD|TX_CLK|TX_CTL|TX_EN|TX_ER|TXD)`

### DIFF
--- a/dts/st/mp1/stm32mp151aaax-pinctrl.dtsi
+++ b/dts/st/mp1/stm32mp151aaax-pinctrl.dtsi
@@ -1504,6 +1504,158 @@
 				slew-rate = "very-high-speed";
 			};
 
+			/* ETH_STM32MP1 */
+
+			/omit-if-no-ref/ eth1_crs_pa0: eth1_crs_pa0 {
+				pinmux = <STM32_PINMUX('A', 0, AF11)>;
+				slew-rate = "very-high-speed";
+			};
+
+			/omit-if-no-ref/ eth1_ref_clk_pa1: eth1_ref_clk_pa1 {
+				pinmux = <STM32_PINMUX('A', 1, AF11)>;
+				slew-rate = "very-high-speed";
+			};
+
+			/omit-if-no-ref/ eth1_rx_clk_pa1: eth1_rx_clk_pa1 {
+				pinmux = <STM32_PINMUX('A', 1, AF11)>;
+				slew-rate = "very-high-speed";
+			};
+
+			/omit-if-no-ref/ eth1_col_pa3: eth1_col_pa3 {
+				pinmux = <STM32_PINMUX('A', 3, AF11)>;
+				slew-rate = "very-high-speed";
+			};
+
+			/omit-if-no-ref/ eth1_crs_dv_pa7: eth1_crs_dv_pa7 {
+				pinmux = <STM32_PINMUX('A', 7, AF11)>;
+				slew-rate = "very-high-speed";
+			};
+
+			/omit-if-no-ref/ eth1_rx_ctl_pa7: eth1_rx_ctl_pa7 {
+				pinmux = <STM32_PINMUX('A', 7, AF11)>;
+				slew-rate = "very-high-speed";
+			};
+
+			/omit-if-no-ref/ eth1_rx_dv_pa7: eth1_rx_dv_pa7 {
+				pinmux = <STM32_PINMUX('A', 7, AF11)>;
+				slew-rate = "very-high-speed";
+			};
+
+			/omit-if-no-ref/ eth1_rxd2_pb0: eth1_rxd2_pb0 {
+				pinmux = <STM32_PINMUX('B', 0, AF11)>;
+				slew-rate = "very-high-speed";
+			};
+
+			/omit-if-no-ref/ eth1_rxd3_pb1: eth1_rxd3_pb1 {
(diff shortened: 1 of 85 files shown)


arifbalik:
I was trying to bring up ethernet on stm32mp1 and realized the pin data is quite different than what was specified in the default yml file, I'll see if I can put it inside the default one but it seems unlikely.

Also last time I updated the pinctrl files @erwango suggested that I should open another PR for the `dtsi` files since they were quite large, i can put it here as well

erwango:
> I updated the pinctrl files @erwango suggested that I should open another PR for the `dtsi`

Maybe I wasn't clear enough, I wanted to mean splitting into disctinct commits, not PRs.

arnopo:
> > > I was trying to bring up ethernet on stm32mp1 and realized the pin data is quite different than what was specified in the default yml file, I'll see if I can put it inside the default one but it seems unlikely.
> > 
> > 
> > perhaps you could just add
> > ```
> > - name: ETH_MP1x
> >   match: "^ETH[12]_(CLK|COL|CRS|CRS_DV|GTX_CLK|MDC|MDIO|PHY_INTN|PPS_OUT|REF_CLK|RX_CLK|RX_CTL|RX_DV|RX_ER|RXD[0-3]|TX_CLK|TX_CTL|TX_EN|TX_ER|TXD[0-3])$"
> >   slew-rate: very-high-speed
> > ```
> 
> The problem is that some of the definitions in the pin data matches some regex expressions inside `stm32-pinctrl-config.yaml` but not all, so if we just add the expression you provided, we would get some definitions twice.
> 
> For example some definitions will match the following regex and this is from `stm32-pinctrl-config.yaml`
> 
> ```yaml
> - name: ETH (Extended)
>   match: "^ETH\\d+_(?:MDC|MDIO|PHY_INTN|PPS_OUT|CLK|TX_ER)$"
>   slew-rate: very-high-speed
> ```
> 
> But none would match against this;
> 
> ```yaml
> - name: ETH
>   match: "^ETH+_(?:COL$|CRS|CRS_DV|MDC|MDIO|PPS_OUT|REF_CLK|RX_CLK|RX_DV|RX_ER||RXD[0-3]|TX_CLK|TX_EN|TXD[0-3])$"
>   slew-rate: very-high-speed
> ```
> 
> Maybe I could remove the expressions inside `EH_MP1x` that causes the double match, but then the pinctrl file would call some ETH pins as "extended" even though they are not.

Is Completing the ` ETH (Extended) ` not sufficient?
Could you provide some examples of names that would not match, explaining why? This would help us better understand the issue.

If we are unable to enhance the stm32-pinctrl-config.yaml, an alternative could be to include the changes in stm32mp1-pinctrl-config.yaml to only override the ETH section.

arnopo:
> > @arifbalik
> > Any update planned? FYI I just pushed in draft the support of the eth1 base you this PR: [zephyrproject-rtos/zephyr#96134](https://github.com/zephyrproject-rtos/zephyr/pull/96134)
> 
> @arnopo I've returned to your email but i guess there was a problem, I was also working with ethernet and had a very similar branch to yours. But it was not ready for upstream.

I did not answer l from you, but I saw that you push some related code on your GitHub. Testing it was not functional.
But I reused a part of your device tree.  So don't hesitate to claim that I add your sign-off on some part of the code I pushed. 

> 
> I havent been able to work on this for a while. But i can get back to it.
> 
> Let me see if i can change the pin config like you suggested. I can get back to you by the end of the week.
Thank!

arifbalik:
> So don't hesitate to claim that I add your sign-off on some part of the code I pushed

That would be great i you could do it

arnopo:
> @arnopo I did managed to put the regex inside the original yaml file. However there is a logical seperation now, because some of the pins match the regex `ETH (Extended)`, and I put the rest in the `ETH_STM32MP1`, so the pins are seperated by `ETH (Extended)` and `ETH_STM32MP1` which may cause confusion.

LGTM but need @erwango feedback

erwango:
@arifbalik Following discussion with @arnopo It is ok to merge mp15 eth pin configurations. The only change would be to change from `ETH_STM32MP13` to `ETH_STM32MP1`

dts/README.rst
```diff
@@ -31,7 +31,7 @@ URL:
    https://github.com/STMicroelectronics/STM32_open_pin_data
 
 Commit:
-   5e641448dc7df619ee46bb0c91796b7626005929
+   3d64b1076ac7f0bf0f74acf89b9578b530f37ed7
```
arnopo: Is there any reason to bump the version? You should generate the files using version 6.15 for the STM32MP.
You should use version 6.14 (5e641448dc7df619ee46bb0c91796b7626005929) for your generation; otherwise, a pinctrl files update for all STM32 platforms would be required according to the new version.

arifbalik: I think this was generated automatically. I will regerate with 6.14

arifbalik: Done

erwango (approved):

arnopo (approved):

PR #292, opened 2025-06-24
hal_stm32 : Fix for SDIO polling mode transfer issues

ExaltZephyr:
Fixes two related issues in the SDIO driver: incorrect block count calculation that caused the card to hang, and improper buffer size handling that led to misaligned FIFO writes. Together, these changes ensure correct and stable data transfer in polling mode.


ExaltZephyr:
> @ExaltZephyr Please rebase to fix the N6/README conflict

Done.

PR #294, opened 2025-07-08
STM32WBAX: Power management rework to support STOP and STDBY modes with radio activity

asm5878:
Power management rework to support STOP and STDBY modes with radio activity


lib/stm32wba/BLE_TransparentMode/Core/Inc/app_conf.h
```diff
@@ -370,7 +370,7 @@ typedef enum
  * System Clock Manager module configuration
  ******************************************************************************/
 
-#define CFG_SCM_SUPPORTED                   (1)
+#define CFG_SCM_SUPPORTED                   (0)
```
etienne-lms, line 373: Is `CFG_SCM_SUPPORTED` still used somewhere? I think it's not. If so, I think it would be worth to remove it definition.

etienne-lms: My bad, it seems used externally...

etienne-lms (approved):

PR #295, opened 2025-07-08
lib: introduction of the VC8000NanoE video encoder library

hfruchet-st:
This PR introduces the VC8000NanoE software stack needed on top of STM32N6 VENC hardware peripheral in order to enable H264 hardware video compression.
This code comes from STM32CubeN6 firmware code located here: https://github.com/STMicroelectronics/STM32CubeN6/tree/main/Middlewares/Third_Party/VideoEncoder/source
Some changes have been made to comply with Zephyr build system and Zephyr kernel defines.
The debug trace system has been adapted and some trace points fixed.
Debug traces are enabled by setting CONFIG_VC8000NANOE_LOG_LEVEL >= LOG_LEVEL_DEBUG.


lib/vc8000nanoe/doc/Hantro.VC8000NanoE.V50x.SW.Integration.Guide-v1.02-20200708.pdf
mathieuchopstm, line 1: Please drop this file - it's large and unneeded. (Also, I'm a bit worried about licensing...)

mathieuchopstm (changes requested):

lib/vc8000nanoe/inc/enccommon.h
```diff
@@ -153,8 +153,12 @@ typedef struct
 
 /* General tools */
 #define ABS(x)          ((x) < (0) ? -(x) : (x))
+#ifndef MAX
 #define MAX(a, b)       ((a) > (b) ?  (a) : (b))
+#endif
+#ifndef MIN
 #define MIN(a, b)       ((a) < (b) ?  (a) : (b))
+#endif
```
erwango: Use of #ifndef is forbidden. See https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html#rule-a-3-macro-name-collisions

erwango (commented):
Please add a README filr documenting lib origin, how it is maintained, potential add on fixes. See https://github.com/zephyrproject-rtos/hal_stm32/blob/main/lib/stm32wba/README.rst as exemple

lib/vc8000nanoe/inc/enccommon.h
erwango: To workaround this, provide and use an alternate definition:
```diff
-#define MAX(a, b)       ((a) > (b) ?  (a) : (b))
+#define STM32_MAX(a, b)       ((a) > (b) ?  (a) : (b))
```

lib/vc8000nanoe/source/h264/H264EncApi.c
erwango, line 1: Sign-off issue on this commit:
```diff
-Signed-off-by: Hugues Fruchet <hugues.fruchet@foss.st.com
+Signed-off-by: Hugues Fruchet <hugues.fruchet@foss.st.com>
```

erwango (changes requested):

lib/vc8000nanoe/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+# Copyright (c) 2025 STMicroelectronics
+#
+# SPDX-License-Identifier: Apache-2.0
+
+zephyr_library_compile_definitions(H1_HAVE_ENCDEBUG_H)
+#zephyr_library_compile_definitions(CONFIG_VENC_DEBUG)
```
mathieuchopstm: If not needed:
```suggestion
```

lib/vc8000nanoe/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+# Copyright (c) 2025 STMicroelectronics
+#
+# SPDX-License-Identifier: Apache-2.0
+
+zephyr_library_compile_definitions(H1_HAVE_ENCDEBUG_H)
```
mathieuchopstm, line 7: I think it would be better to have a `zephyr_library()` call at the top.
```suggestion

zephyr_library()

zephyr_library_compile_definitions(H1_HAVE_ENCDEBUG_H)
```

lib/vc8000nanoe/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+#zephyr_library_compile_definitions(CONFIG_VENC_DEBUG)
+zephyr_include_directories(inc)
+zephyr_include_directories(source/common)
+zephyr_include_directories(source/h264)
```
mathieuchopstm, line 10: Can we scope (some of) this to library level (`zephyr_library_include_directories`), or are all definitions needed globally because used by the driver?

lib/vc8000nanoe/README.rst
```diff
@@ -0,0 +1,106 @@
+STM32N6 VC8000NanoE hardware video encoder library
+##################################################
+
+Origin:
+   ST Microelectronics
+   https://github.com/STMicroelectronics/STM32CubeN6/tree/main/Middlewares/Third_Party/VideoEncoder/source
```
mathieuchopstm:
```suggestion
   https://github.com/STMicroelectronics/STM32CubeN6/tree/main/Middlewares/Third_Party/VideoEncoder/
```

lib/vc8000nanoe/README.rst
```diff
@@ -0,0 +1,106 @@
...
+Maintained-by:
+   External
+
+License:
+   BSD-3-Clause
+
+License Link:
+   opensource.org/licenses/BSD-3-Clause
+
+Patch List:
```
mathieuchopstm: I know that's not the practice we've had until now, but we discussed possibly providing patches as dedicates files somewhere in here, or at least having a standard way to indicate *where we did the patches* - maybe now's the time to do this, because the patch list here seems quite heavy. cc @erwango

lib/vc8000nanoe/source/h264/H264CodeFrame.c
```diff
@@ -0,0 +1,897 @@
...
+    else if ((inst->numViews > 1) && (inst->numRefBuffsLum == 1) &&
+             (inst->slice.frameNum % 2) && (inst->slice.frameNum > 1))
+        regs->frameCodingType = ASIC_MVC_REF_MOD;
+    else if ((inst->numViews > 1) && (inst->slice.frameNum % 2))
+        regs->frameCodingType = ASIC_MVC;
+    else
+        regs->frameCodingType = ASIC_INTER;
+
+    qpDeltaGainFloat = (float)inst->rateControl.rcMSESum / inst->rateControl.mbPerPic;
+    //printf("qpDeltaGainFloat=%f\n",qpDeltaGainFloat);
```
mathieuchopstm, line 464: debugging code leftovers?
```suggestion
```
ditto the few other such `printf`s here

lib/vc8000nanoe/README.rst
```diff
@@ -0,0 +1,106 @@
...
+
+	* Add zephyr build system
+	  Impacted files: CMakeLists.txt
+                          README.rst
```
mathieuchopstm: I think we only list modifications of the original software in this section(?)
```suggestion
```

stm32cube/stm32n6xx/drivers/src/stm32n6xx_hal_dcmipp.c
```diff
@@ -1041,7 +1041,8 @@ HAL_StatusTypeDef HAL_DCMIPP_PIPE_SetConfig(DCMIPP_HandleTypeDef *hdcmipp, uint3
 
   if (hdcmipp->State == HAL_DCMIPP_STATE_READY)
   {
-    if ((pipe_state == HAL_DCMIPP_PIPE_STATE_RESET) || (pipe_state == HAL_DCMIPP_PIPE_STATE_ERROR))
+    if ((pipe_state == HAL_DCMIPP_PIPE_STATE_READY) || (pipe_state == HAL_DCMIPP_PIPE_STATE_RESET) ||
+        (pipe_state == HAL_DCMIPP_PIPE_STATE_ERROR))
```
mathieuchopstm, line 1045: If this is an intended HAL patch, it should be added to the patch list of the HAL `README`. Otherwise, please drop the change.

lib/vc8000nanoe/README.rst
mathieuchopstm: Regardless of patch file inclusion - the standard practice so far is to `#if 0`-out code (+ personally, add an explicit comment) rather than straight up deleting stuff. See for example: https://github.com/zephyrproject-rtos/hal_stm32/blob/76d62eedec9638254d046d31d56d141c6429dd4d/stm32cube/stm32wb0x/soc/system_stm32wb0x.h#L204-L206

hfruchet-st: This is what we did in the "good old days"... before git !
Now git is there and each changes done on top of initial software is identified through a dedicated commit, so there is no need to track changes inside the source code itselft and make it even more heavier than it is ...

lib/vc8000nanoe/source/h264/H264CodeFrame.c
hfruchet-st: Modifications of initial software stack is limited to absolutely needed functional changes.

lib/vc8000nanoe/CMakeLists.txt
hfruchet-st: The 3 include directories are needed, some .h are in source/ and h264/

lib/vc8000nanoe/README.rst
hfruchet-st: Ok dropped.

lib/vc8000nanoe/README.rst
hfruchet-st: sync'ed with Erwan, I will do that only for removed code.

mathieuchopstm: > This is what we did in the "good old days"... before git ! Now git is there and each changes done on top of initial software is identified through a dedicated commit, so there is no need to track changes inside the source code itselft and make it even more heavier than it is ...

Right, I did not check history and notice you did these removals in dedicated commits :man_facepalming: In this case, LGTM.

I was used to Cube introduction which is done in this repo w/o this "one commit for files, one commit for patches"...

mathieuchopstm (approved):

lib/vc8000nanoe/inc/encdebug.h
```diff
@@ -0,0 +1,69 @@
...
+#include <zephyr/kernel.h>
+
+#ifdef CONFIG_VENC_DEBUG
+#define ASSERT(expr) __ASSERT_NO_MSG(expr)
+#define DEBUG_PRINT(args) printf(args);printf("\n");
+#define COMMENT(x) printf(x);printf("\n");
+#define TRACE_BIT_STREAM(v,n)
+#define APITRACE(x) printf(x);printf("\n");
+#define APITRACEPARAM(x,y) printf("  %s: %d\n",x,y);
+#define APITRACEPARAMX(x,y) printf("  %s: 0x%x\n",x,y);
```
mathieuchopstm: I would use the `LOG`-related infrastructure rather than `printf`.

lib/vc8000nanoe/inc/encdebug.h
```diff
@@ -0,0 +1,69 @@
...
+ *  Description : Encoder debug definitions
+ *
+ ********************************************************************************
+ */
+
+#ifndef __ENC_DEBUG_H__
+#define __ENC_DEBUG_H__
+
+#define CONFIG_ASSERT
+#define CONFIG_ASSERT_LEVEL 2
```
mathieuchopstm: That feels hacky.

lib/vc8000nanoe/inc/encdebug.h
```diff
@@ -0,0 +1,69 @@
...
+
+#include <zephyr/kernel.h>
```
mathieuchopstm: cc @erwango on whether we allow reverse dependency

lib/vc8000nanoe/inc/encdebug.h
hfruchet-st: I'm not fluent on zephyr assertion infrastructure but if those two defines are not set the "__ASSERT_NO_MSG(expr)" is an empty macro and nothing at all is asserted... which is dramatic for debug purpose.
So please advise on how to enable assertions in "not hacky" mode.

lib/vc8000nanoe/inc/encdebug.h
hfruchet-st: This is linked to assertion comment, __ASSERT_NO_MSG(expr) requires this inclusion, if not correct please advise on how to do that.

lib/vc8000nanoe/inc/encdebug.h
mathieuchopstm: I think you're just supposed to enable it using Kconfig (i.e., `west build ... -- -DCONFIG_ASSERT=y`). Or are you saying the library does not see Kconfig symbols as `CONFIG_xxx` defines?

lib/vc8000nanoe/inc/encdebug.h
mathieuchopstm: Yes, I understand what this is for; but on paper, code in HAL is """""not allowed to call back to Zephyr""""". This is a very specific use case though so IMO it's fine, but I'll let Erwan decide.

lib/vc8000nanoe/inc/encdebug.h
```diff
@@ -0,0 +1,69 @@
...
+
+#ifdef CONFIG_VENC_DEBUG
```
mathieuchopstm: On that note, this should be an actual Kconfig on the Zephyr side and not merely something you `-DCONFIG_VENC_DEBUG` as part of CMake commandline. (though if we move to `LOG`ging, the question solves itself)

lib/vc8000nanoe/inc/encdebug.h
erwango: `h264encapi.h` defines `void H264EncTrace(const char *msg);`
What about implementing this function `drivers/video/video_stm32_venc.c` with as much zephyr api and trace control as required ?
Note that it can be left for another PR (and current commit dropped).

lib/vc8000nanoe/inc/encdebug.h
```diff
@@ -0,0 +1,74 @@
...
+#define __ENC_DEBUG_H__
+
+#ifdef CONFIG_ASSERT
+#define STR(x) #x
+#define ASSERT(expr)                                                      \
+	do {                                                              \
+		if (!(expr)) {                                            \
+			EWLassert((bool)(expr), STR(expr), __FILE__, __LINE__);   \
+		}                                                         \
+	} while (false)
```
mathieuchopstm, line 53: Is `if (!(expr))` added as an "optimization" (to avoid calling `EWLassert` if everything is fine)? Otherwise, I don't see the point of providing `expr` to `EWLassert`, since it's only called when the assertion check failed.

lib/vc8000nanoe/inc/basetype.h
```diff
@@ -104,13 +104,14 @@ typedef int32_t i32;
 typedef uint64_t u64;
 typedef int64_t   i64;
 
-
+#ifndef true
```
mathieuchopstm: Nit: I'd add a `/** PATCH */` comment - it's nice to have the explanation directly instead of having to look at the Git history :slightly_smiling_face:

mathieuchopstm (approved):
Two minor points, LGTM otherwise.

lib/vc8000nanoe/inc/encdebug.h
hfruchet-st: Yes, it is redundant with check made itself by back-end of EWLassert, but better to not call a function to do this check...

lib/vc8000nanoe/inc/basetype.h
hfruchet-st: Sorry, I'm not keen of that, better to not overload code with something that git solve.

lib/vc8000nanoe/inc/enccommon.h
```diff
@@ -153,8 +153,12 @@ typedef struct
 
 /* General tools */
 #define ABS(x)          ((x) < (0) ? -(x) : (x))
+#ifndef MAX
```
hfruchet-st: @erwango , I know we discussed about that, but I have no other way to fix that in order to fix all warnings:
on one hand the library must not call zephyr ".h", so I cannot include "zephyr/utils.h" to solve MIN/MAX definitions and on the other hand zephyr driver video_stm32_venc.c indirectly includes "zephyr/utils.h", and also indirectly includes "enccommon.h" through "h264encapi.h", so I need to prevent multiple definitions anyway...

erwango (approved):

PR #300, opened 2025-08-01
stm32cube: stm32fx: remove phy init and config 

maass-hamburg:
remove phy init and config from the legacy ethernet hal,
as that is already done in zephyr before and we don't want it to
be overwritten


juickar (approved):
These changes look good to me
Tested on stm32f207zg with the dhcpv4 client sample

PR #302, opened 2025-08-06
Add stm32 COMP pins

KozhinovAlexander:
add comparator pins config


KozhinovAlexander:
@erwango @mathieuchopstm - could you please take a look and assign a right reviewer?

PR #303, opened 2025-08-12
lib: stm32wb0: Provide support for using radio timer as the system clock

HoZHel:
Provide support for using the radio timer as the system clock when CONFIG_STM32_RADIO_TIMER is set.

Transfer BLEPLAT_SetRadioTimerValue function to the Bluetooth HCI driver in Zephyr.

Supplementary PR [#94402](https://github.com/zephyrproject-rtos/zephyr/pull/94402)


lib/CMakeLists.txt
```diff
@@ -8,5 +8,7 @@
 if(CONFIG_HAS_STM32LIB)
   add_subdirectory_ifdef(CONFIG_BT_STM32_IPM stm32wb)
   add_subdirectory_ifdef(CONFIG_BT_STM32WBA stm32wba)
-  add_subdirectory_ifdef(CONFIG_BT_STM32WB0 stm32wb0)
+  if(CONFIG_BT_STM32WB0 OR CONFIG_STM32_RADIO_TIMER)
```
asm5878: Implicit assumption here is that radio timer is only available on wb0.
You could also check if  we are in wb0.
But it's really a detail.

asm5878 (approved):

erwango:
`DNM` untill https://github.com/zephyrproject-rtos/zephyr/pull/94402 is approved

lib/CMakeLists.txt
HoZHel: Do you mean as an extra check? Or instead of this line?

mathieuchopstm: I had the same thought. Would be fine if the Kconfig symbol was WB0-tainted (e.g., `CONFIG_STM32WB0_RADIO_TIMER`)

HoZHel: Done.

HoZHel:
Rebased to main to resolve a conflict.

etienne-lms (approved):

PR #304, opened 2025-08-13
dcmipp: allow PIPE_SetConfig when pipe state is READY

avolmat-st:
Allow to call HAL_DCMIPP_PIPE_SetConfig when the pipe state is READY. Currently it is only possible to use this function if the pipe state is RESET and ERROR states.
Indeed, it is possible to reconfigure the PIPE as soon as the pipe is not currently being used, aka in the READY state. With that done, it becomes possible to change the PIPE configuration between two use-cases without having to DeInit / Init the HAL_DCMIPP or use the unitary pixel packer functions.

### DIFF
--- a/stm32cube/stm32h7rsxx/README
+++ b/stm32cube/stm32h7rsxx/README
@@ -61,4 +61,13 @@ Patch List:
      Impacted file: 
       stm32cube/stm32h7rsxx/drivers/src/stm32h7rsxx_hal_sdio.c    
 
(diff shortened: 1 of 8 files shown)


erwango:
Please push a Zephyr PR targeting this change (aim is to verify merging it won't break anything).

avolmat-st:
> Please push a Zephyr PR targeting this change (aim is to verify merging it won't break anything).

Push https://github.com/zephyrproject-rtos/zephyr/pull/95368 for testing.

PR #306, opened 2025-08-13
hal_stm32: dts:  h7/r/s :  add missing eth_mii and eth_rmii signals 

djiatsaf-st:
This PR updates the h7r/s-pinctrl.dtsi by adding missing `eth_mii` and `eth_rmii` signals.

fixes issue https://github.com/zephyrproject-rtos/zephyr/issues/94268

### DIFF
--- a/scripts/genpinctrl/stm32-pinctrl-config.yaml
+++ b/scripts/genpinctrl/stm32-pinctrl-config.yaml
@@ -77,7 +77,7 @@
   slew-rate: very-high-speed
 
 - name: ETH_MII
-  match: "^ETH\\d_MII_(?:COL|CRS|RXD[0-3]|RX_CLK|RX_DV|RX_ER|TX_EN|TXD[0-3]|TX_CLK|TX_EN)$"
+  match: '^ETH\d?_MII_(?:COL|CRS|RXD[0-3]|RX_CLK|RX_DV|RX_ER|TX_EN|TXD[0-3]|TX_CLK|TX_EN)$'
   slew-rate: very-high-speed
 
 - name: ETH_RGMII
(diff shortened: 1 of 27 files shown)


scripts/genpinctrl/stm32-pinctrl-config.yaml
```diff
@@ -77,15 +77,15 @@
...
-  match: "^ETH\\d_MII_(?:COL|CRS|RXD[0-3]|RX_CLK|RX_DV|RX_ER|TX_EN|TXD[0-3]|TX_CLK|TX_EN)$"
+  match: '^ETH\d?_MII_(?:COL|CRS|RXD[0-3]|RX_CLK|RX_DV|RX_ER|TX_EN|TXD[0-3]|TX_CLK|TX_EN)$'
   slew-rate: very-high-speed
 
 - name: ETH_RGMII
   match: "^ETH\\d+_RGMII_(?:CLK125|GTX_CLK|RXD[0-3]|RX_CLK|RX_CTL|RX_ER|TX_EN|RX_DV|TXD[0-3]|TX_CTL|TX_EN)$"
   slew-rate: very-high-speed
 
 - name: ETH_RMII
-  match: "^ETH\\d+_RMII_(?:CRS_DV|REF_CLK|RXD[0-1]|RX_CLK|RX_ER|TX_EN|RX_DV|TXD[0-1]|TX_CLK|TX_EN)$"
```
erwango, line 88: It seems that `CRS_DV|REF_CLK` removal is the difference with https://github.com/zephyrproject-rtos/hal_stm32/pull/306#issuecomment-3216609242, which appears to be right in this regard.

However, I do prefer conciseness of this version.
@mariopaja If you have free cycles, can you adapt your PR to this style ?
Otherwise, we'll wait @djiatsaf-st is back from holidays to fix this.

mariopaja: @erwango then should I remove `CRS_DV|REF_CLK` from all affected series?

erwango: No, I mean these should be present and you're PR is right in that regard.
But, I'd prefer you could adopt the same conciseness and keeping the 2 filters instead of adding 2 more.

PR #307, opened 2025-08-18
stm32cube: stm32h7xx: drivers: Correct DMA source and destination addresses

kbidani:
The source and destination addresses in the HAL_MDMA_Start_IT call were swapped, causing incorrect data transfers. The data buffer (pData) should be the source, and the OSPI data register (DR) should be the destination. This commit fixes the argument order to ensure proper data transmission via DMA.

fixes: #94350

### DIFF
--- a/stm32cube/stm32h7xx/README
+++ b/stm32cube/stm32h7xx/README
@@ -75,4 +75,9 @@ Patch List:
      Impacted file: 
       stm32cube/stm32h7xx/drivers/src/stm32h7xx_hal_sdio.c
          
+   *Fix an issue in the STM32H7 HAL OSPI by correcting the DMA source and destination
+      addresses in the HAL_MDMA_Start_IT() call.
+      Impacted files:
+      stm32cube/stm32h7xx/drivers/src/stm32h7xx_hal_ospi.c
+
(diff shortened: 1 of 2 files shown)


stm32cube/stm32h7xx/README
```diff
@@ -68,4 +68,9 @@ Patch List:
      Impacted files:
       stm32cube/stm32h7xx/drivers/src/stm32h7xx_hal_sdio.c
 
+   *Fix an issue in the STM32H7 HAL OSPI by correcting the DMA source and destination
+      addresses in the HAL_MDMA_Start_IT() call.
+      Impacted files:
+      stm32cube/stm32h7xx/drivers/src/stm32h7xx_hal_ospi.c
```
erwango, line 81: Is there an internal issue associated to be sure this is fixed directly in HAL ?

kbidani: The bug exists on the HAL side but their reference does not point to the latest cube version V1.12.0 and HAL tag 1.11.5 where the bug is present.

erwango (approved):

FRASTM (approved):

PR #308, opened 2025-08-25
Add stm32 opamp pins

KozhinovAlexander:
Add stm32 opamp pins

### DIFF
--- a/scripts/genpinctrl/stm32-pinctrl-config.yaml
+++ b/scripts/genpinctrl/stm32-pinctrl-config.yaml
@@ -167,6 +167,9 @@
   match: "^OCTOSPI(.*)(?:CLK|NCS|DQS|IO[0-7])$"
   slew-rate: very-high-speed
 
+- name: OPAMP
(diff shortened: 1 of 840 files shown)


scripts/genpinctrl/stm32-pinctrl-config.yaml
```diff
@@ -167,6 +167,12 @@
   match: "^OCTOSPI(.*)(?:CLK|NCS|DQS|IO[0-7])$"
   slew-rate: very-high-speed
 
+- name: OPAMP_VIN
+  match: "^OPAMP\\d+_VIN[MP]$"
+
+- name: OPAMP_VOUT
+  match: "^OPAMP\\d+_VOUT$"
```
erwango: Can you make it a one liner ? Having multiple lines makes generation longer.
(We have clean up to do on other IPs as well)

erwango (changes requested):

scripts/genpinctrl/stm32-pinctrl-config.yaml
KozhinovAlexander: done

PR #312, opened 2025-09-29
Made it possible to use the build in hardware support for clause 45 mdio communication

GustavSVJ:
I am using the STM32H7 along with Zephyr and a TJA1103 phy. The phy requires Clause 45 support which the STM32H7 has but it is not included in the hal layer and therefore not available in Zephyr.

I've tried opening an issue in the ST repo but it was denied due to no development boards requiring the functionoality. Another customer also got the same response on a PR a while ago.

So now I'm hoping for the next best solution...

[ST PR](https://github.com/STMicroelectronics/stm32h7xx-hal-driver/pull/52)
[ST Issue](https://github.com/STMicroelectronics/stm32h7xx-hal-driver/issues/90)

### DIFF
--- a/stm32cube/stm32h7xx/drivers/include/stm32h7xx_hal_eth.h
+++ b/stm32cube/stm32h7xx/drivers/include/stm32h7xx_hal_eth.h
@@ -1767,6 +1767,10 @@ HAL_StatusTypeDef HAL_ETH_WritePHYRegister(const ETH_HandleTypeDef *heth, uint32
                                            uint32_t RegValue);
 HAL_StatusTypeDef HAL_ETH_ReadPHYRegister(ETH_HandleTypeDef *heth, uint32_t PHYAddr, uint32_t PHYReg,
                                           uint32_t *pRegValue);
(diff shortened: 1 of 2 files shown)


erwango:
@GustavSVJ As you mentioned this fonctionality won't be implemented on H7 Cube for specific reasons.
So rather than updating HAL, can it be implemented directly within Zephyr driver ?
This change seems rather self contained.

GustavSVJ:
Dear @erwango,

Would you implement the full function in the zephyr driver or would you modify the specific bits and then call the existing "HAL_ETH_WritePHYRegister" read function?

Both solutions seem wrong to me. Re-implementing the full function is going in the opposite direction of the rest of the driver which relies on the ST HAL layer. Modifying the specific bits also seems dangerous as any changes to the ST HAL layer underneath could break the functionality in a way only showing up at runtime.

erwango:
@GustavSVJ As an alternative, I propose you have a look to DW driver which could also be used on H7 and on which the implementation you request should be easier to add if not already present (https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/ethernet/eth_dwmac.c).
Meanwhile, I'm closing this issue.

PR #321, opened 2025-10-22
stm32cube: get rid of the HAL Legacy definitions header

mathieuchopstm:
Remove the HAL Legacy definitions header from all series using a little trick to ensure proper compilation despite the missing file, and update the README of all series to remove mention of patches to the header we are getting rid of.

Also remove the other legacy headers (can_legacy / eth_legacy) which are never used from the Zephyr side. This is however more of a cosmectic patch as these headers are not part of the build (they are only included if the corresponding HAL module is enabled, which is not the case in Zephyr).


etienne-lms (approved):

djiatsaf-st (approved):

erwango (approved):
Approved. Taking into account that this should be taken into account in upcoming HAL update script refactoring (before next HAL update).

mathieuchopstm:
Moving to draft/DNM: Zephyr CI revealed there are still drivers we need that are "buggy".

I'll resume work on this after properly assessing how many HAL drivers are dependent on Legacy definitions, rather than based only on smoke tests as I've done so far...

PR #335, opened 2025-12-16
scripts: implement automatic generation of various files

mathieuchopstm:
Add a new script `autogenpackage` (better name suggestions welcome) which can be used to automatically generate various files that were so far hand-authored:
* list of series supported by the HAL module
* HAL assert header `stm32_assert.h`
  * Build system is slightly tweaked to use a common file as done with `common_ll`
  * This makes the package more maintainable and reduces space usage
* per-series `CMakeLists.txt`

Additionally, the script can generate the Zephyr-side `modules/Kconfig.stm32`, ensuring that the HAL and Zephyr side of the module are always in sync. Other features include the ability to gate different drivers under the same Kconfig option (e.g., `DLYB` for drivers called `dlyb.c` on some series and `delayblock.c` on others) and easily include or force-exclude drivers from the build system (although this should be done as little as possible).

As part of this PR, the series list and `stm32_assert.h` are updated for the new infrastructure, but per-series `CMakeLists.txt` is left unchanged for now.

There are still some points to discuss (e.g., script is oriented towards `system_stm32XXxx.c` removal) so this PR is opened as draft for discussion.


djiatsaf-st (commented):
I tested locally, and it works `fine`, but I have some questions /suggestions:

- Is it okay to remove these copyrights?

    https://github.com/zephyrproject-rtos/hal_stm32/blob/802536378244fbed6b560c7e74d8ffa1e010a3f2/stm32cube/stm32g0xx/CMakeLists.txt#L1
     https://github.com/zephyrproject-rtos/hal_stm32/blob/802536378244fbed6b560c7e74d8ffa1e010a3f2/stm32cube/stm32wbxx/CMakeLists.txt#L1
- Is it possible to add a `--only-series` option for a limited scope? This could be used when introducing a new series to generate packages only for that series instead of all series.

mathieuchopstm:
> * Is it okay to remove these copyrights?
>   https://github.com/zephyrproject-rtos/hal_stm32/blob/802536378244fbed6b560c7e74d8ffa1e010a3f2/stm32cube/stm32g0xx/CMakeLists.txt#L1

The new files are entirely generated from scratch (i.e., they have no relation to the "old" ones) so I don't think the old copyright has any reason to apply. On the other hand, maybe the copyright notice should be removed from the script-generated headers since they are merely script-generated.

>   https://github.com/zephyrproject-rtos/hal_stm32/blob/802536378244fbed6b560c7e74d8ffa1e010a3f2/stm32cube/stm32wbxx/CMakeLists.txt#L1
> * Is it possible to add a `--only-series` option for a limited scope? This could be used when introducing a new series to generate packages only for that series instead of all series.

Why? The script is idempotent, if you run it multiple times on the same files (i.e., unmodified series) then it won't have any effect. Unlike `genpinctrl`, there's also no performance issue here so I don't see the point of implementing filtering. Also, some operations of the script are global (based on all series) so how would filtering even work? Ignore the excluded series entirely? Ignore them individually but include them for global?

mathieuchopstm:
Will be obsoleted by new script (not released yet) - closing.

PR #353, opened 2026-03-16
ci: bump test matrix to Python 3.12 and newer

mathieuchopstm:
Zephyr now requires Python 3.12 as minimum version (see https://github.com/zephyrproject-rtos/zephyr/pull/104317).

Drop Python 3.10 and 3.11 from test matrix and, while at it, add 3.13 and 3.14 to catch forward compatibility issues early on (though Python 3.12 EOL is [October 2028](https://devguide.python.org/versions/) so we still have lots of time).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,7 +30,7 @@ jobs:
             --cov-report term \
             --cov-report html \
             -vv
-      - name: store coverage report 
+      - name: store coverage report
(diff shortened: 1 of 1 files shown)


kartben (approved):

str4t0m (approved):

mathieuchopstm:
@erwango I think we can proceed with this now?

erwango (approved):

PR #358, opened 2026-03-24
stm32cube: Fix License scheme for non HAL2 based series

erwango:
For all STM32 series that are not based on HAL2, update the license scheme to better reflect their original license.
So all "CMSIS" based files are now under Apache2.0 instead of BSD-3 Clause.

Apache 2.0 license file is added in stm32cube/stm32*/soc/ folders BSD-3 Clause license file is moved under stm32cube/stm32*/drivers/ folders.


etienne-lms (approved):
All consistent.

mathieuchopstm (approved):
Example on STM32CubeH5 package (indicative only, for future readers):
* https://github.com/STMicroelectronics/STM32CubeH5/blob/e25dd1b1d0e8871627e79e2d1453687f0f22835c/Drivers/CMSIS/LICENSE.txt
* https://github.com/STMicroelectronics/cmsis-device-h5/blob/4a9d636e98543d21553ea70a3925f791f62afe34/LICENSE.md

juickar (approved):

gautierg-st (approved):

PR #367, opened 2026-05-18
Re-generate pinctrl files

etienne-lms:
Update generated pinctrl DTSI files that was wrongly updated by commit de30f8721b26 that was rebased without re-generating the DTSI file while previously merged commit 1622efd17c8d0 added generation of pinctrl nodes related to I2S, JTAG TRACECLK, USB SOF for HAL1 based SoCs as well as a modification of the XSPIM pinctrl nodes heading inline comment.

Prevent Git from tracking removal on pinctrl sub-directories for SoCs relying on HAL2 (currently only the stm32c5 SoC family) when using **genpinctrl.py** scirpt.

### DIFF
--- a/scripts/genpinctrl/genpinctrl.py
+++ b/scripts/genpinctrl/genpinctrl.py
@@ -17,7 +17,7 @@ import logging
 from pathlib import Path
 import re
 import shutil
-from subprocess import check_output, STDOUT, CalledProcessError
+from subprocess import check_output, STDOUT, CalledProcessError, run
 import xml.etree.ElementTree as ET
 
 from jinja2 import Environment, FileSystemLoader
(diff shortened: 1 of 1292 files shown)


rgallaispou (approved):
LGTM with the try except.

scripts/genpinctrl/genpinctrl.py
```diff
@@ -78,6 +78,11 @@
 ]
 """Supported SoC families"""
 
+HAL2_FAMILIES = [
+    "stm32c5",
+]
+"""SoC families relying on HAL2"""
```
erwango, line 84: Ok for now, but we'll need a way to make this more generic. @mathieuchopstm can you check that ?

erwango (approved):

scripts/genpinctrl/genpinctrl.py
mathieuchopstm: see #369

PR #370, opened 2026-05-21
stm32cube: move series list from CMakeLists.txt to a dedicated file

mathieuchopstm:
This may appear useless for now but will come in handy later as it makes automatic generation of the list almost trivial.


erwango (approved):

rgallaispou (approved):
LGTM.

PR #376, opened 2026-06-10
scripts: genllheaders: HAL2 compatibility + generate READMEs with final newline

mathieuchopstm:
Allow `genllheaders` to run now that there are STM32Cube HAL2 series in tree (I wonder how STM32CubeC5 support was introduced in the first place considering how the script is broken...)

While at it, also modify the script to generate README files with a newline at the end, since that's how all the files are in tree and makes more sense anyways.


djiatsaf-st (approved):

gautierg-st (approved):
To answer your question about C5, I started out the development following the directory structure of HAL1 (so genllheaders worked) and later changed to follow the native HAL2 structure (while keeping the previously generated files and without trying to regenerate them).

erwango (approved):

PR #383, opened 2026-06-23
RFC: Add a -pinctrl.dtsi generation validation action

### LINKED ISSUE #109921: hal_stm32: Action to validate -pinctrl.dtsi state
### Summary

*-pinctrl.dtsi files in hal_stm32 repo are generated based on a repos (https://github.com/STMicroelectronics/STM32_open_pin_data for HAL1 version, DFP pack for HAL2 versions) and specific filters that are modified each time signals for a new HW IP are needed, or new versions of these packs arizes.

Combination of these (pack versions + state of the filter) results is the new batch of the *-pinctrl.dtsi files.
While we review the changes to these carefully, it is still possible that errors fall through the cracks resulting to a non aligned combination of version/filter/batch making it harder for the next one to modify.

We miss an automated check to verify the state of these files before merging a new PR.

### Describe the solution you'd like

Add a action that will automatically run the generation script on the new branch on top of each PR modifying these files.
Check is easy: if no changes are seen, on top of the branch, the PR is correct. Otherwise there is an error (last SHA1 not correctly documented, partial generation, wrong filter, ...)

### Alternatives

_No response_

### Additional Context

_No response_

erwango:
Aim of the action is to validate state of the -pinctrl.dtsi files if they are modified.

This should provide an easy way for maintainer to verify that pinctrl.dtsi files have been modified and documented according to the rules. It also provides a quick feedback to contributors on their changes

The action is triggered in the PR touching these files, config files or README files that document the versions of the input components.
When triggered it  will generate a new set of -pcintrl.dtsi files based on the documented versions of STM32_open_pin_data (HAL1) and DFP (HAL2). After generation, it checks that there is no difference with the commited version.
If differences are found, an error is generated.

The script is also able to take into account patches documented in README files, and in that case, it only checks that a difference is available in the documented files, but isn't able to verify the nature of the difference.

Since the script is looking for information (repo versions, patched files) in the README files that are not strictly structured, there is a fragility on this part. TBD if acceptable or if we should move to another way to document those in a more binding way (.yaml file for instance), which would be a more impacting change.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,6 +3,232 @@ name: Test
 on: [push, pull_request]
 
 jobs:
+  pinctrl-validation:
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: check if pinctrl files were modified
+        id: changed
+        run: |
+          # Fetch the base commit so git diff can compare against it.
+          git fetch --no-tags --depth=1 origin "${{ github.event.pull_request.base.sha }}"
+
+          # Determine whether any pinctrl-relevant path changed:
+          #   - dts/**/*.dtsi      : generated pinctrl files
+          #   - dts/README.rst     : documents the HAL1 (STM32_open_pin_data) revision
+          #   - scripts/genpinctrl : the generation scripts themselves
+          pinctrl_changed=false
+          if ! git diff --quiet "${{ github.event.pull_request.base.sha }}" HEAD -- \
+            ':(glob)dts/**/*.dtsi' \
+            'dts/README.rst' \
+            'scripts/genpinctrl'; then
+            pinctrl_changed=true
+          fi
+
+          # For HAL2-based series (those with a dfp/ subdirectory), a bump of the
+          # series README also requires regenerating the pinctrl files.
+          # HAL1-based series READMEs are unrelated to pinctrl generation and must
+          # not trigger this job unnecessarily.
+          if [ "$pinctrl_changed" = false ]; then
+            for readme in $(git diff --name-only \
+                "${{ github.event.pull_request.base.sha }}" HEAD \
+                -- ':(glob)stm32cube/*/README'); do
+              series_dir=$(dirname "$readme")
+              if [ -d "$series_dir/dfp" ]; then
+                pinctrl_changed=true
+                break
+              fi
+            done
+          fi
+
+          echo "pinctrl=$pinctrl_changed" >> "$GITHUB_OUTPUT"
+
+      - name: Set up Python
+        if: steps.changed.outputs.pinctrl == 'true'
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - name: install dependencies
+        if: steps.changed.outputs.pinctrl == 'true'
+        run: |
+          pip3 install -r scripts/requirements.txt
+
+      - name: get expected STM32 Open Pin Data commit
+        if: steps.changed.outputs.pinctrl == 'true'
+        id: opendata
+        run: |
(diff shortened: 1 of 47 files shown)


.github/workflows/test.yml
```diff
@@ -3,6 +3,185 @@ name: Test
...
+            if [ -z "$cube_url" ] || [ -z "$cube_commit" ]; then
+              echo "Failed to find STM32Cube URL or commit in $cube_readme"
+              exit 1
+            fi
+
+            cube_repo=${cube_url#https://github.com/}
+            cube_repo=${cube_repo%.git}
+            api_url="https://api.github.com/repos/$cube_repo/contents/$dfp_path?ref=$cube_commit"
+            api_header="Accept: application/vnd.github+json"
+            curl -fsSL -H "$api_header" "$api_url" > /tmp/dfp.json
```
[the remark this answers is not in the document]
erwango: This change was made on purpose to fit our internal gh org constraints.
I don't think that low anonymous rate limit would be an issue for the purpose of this usage, but I can have a try this way.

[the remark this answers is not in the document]
mathieuchopstm: `${{ secrets.GITHUB_TOKEN }}` - rather than `${{ github.token }}` - is mentioned in the documentation for all flavors of GitHub ([Web](https://docs.github.com/en/actions/tutorials/authenticate-with-github_token), [GHE Server](https://docs.github.com/en/enterprise-server@3.21/actions/tutorials/authenticate-with-github_token) and [GHE Cloud](https://docs.github.com/en/enterprise-cloud@latest/actions/tutorials/authenticate-with-github_token)) so I believe it should work internally too.

.github/workflows/test.yml
```diff
@@ -3,6 +3,185 @@ name: Test
...
+
+            dfp_url=$(python3 -c 'import json, sys; print(json.load(open(sys.argv[1])).get("submodule_git_url", ""))' /tmp/dfp.json)
+            dfp_commit=$(python3 -c 'import json, sys; print(json.load(open(sys.argv[1])).get("sha", ""))' /tmp/dfp.json)
+            dfp_type=$(python3 -c 'import json, sys; print(json.load(open(sys.argv[1])).get("type", ""))' /tmp/dfp.json)
```
mathieuchopstm: Nit: perhaps it is possible to invoke Python only once, printing `submodule_git_url | sha | type` then splitting at shell script level?

mathieuchopstm (commented):
Seems fine otherwise

erwango:
@mathieuchopstm PTAL and I'll remove the test commit.

.github/workflows/test.yml
```diff
@@ -3,6 +3,232 @@ name: Test
...
+
+      - name: regenerate pinctrl files (HAL1)
+        if: steps.changed.outputs.pinctrl == 'true'
+        run: |
+          python3 scripts/genpinctrl/genpinctrl.py -p /tmp/STM32_open_pin_data -o dts
+
+      - name: regenerate pinctrl files (HAL2)
+        if: steps.changed.outputs.pinctrl == 'true'
+        run: |
+          python3 scripts/genpinctrl/genpinctrl_hal2.py -p /tmp/STM32Cube_DFP -o dts
```
mathieuchopstm, line 140: It could be sensible to move at least all of this, and perhaps even the patches check, to a dedicated Shell script such that it can be also used manually to perform regeneration.

Can be done at a later time though.

mathieuchopstm (approved):

.github/workflows/test.yml
erwango: I indeed thought about it, but I was hesitant due to the git manipulation. Let's do it afterwards.

erwango:
Thanks for the approval. Removing the test commit.

PR #389, opened 2026-07-08
dts: st: add LCD related pinctrl

avolmat-st:
Update the stm32-pinctrl-config.yaml to also generate LCD related pinctrls and generate all of them.

### DIFF
--- a/dts/st/l4/stm32l476v(c-e-g)tx-pinctrl.dtsi
+++ b/dts/st/l4/stm32l476v(c-e-g)tx-pinctrl.dtsi
@@ -1068,6 +1068,232 @@
 		pinmux = <STM32_PINMUX('E', 6, AF0)>;
 	};
 
+	/* LCD */
+
+	/omit-if-no-ref/ lcd_seg0_pa1: lcd_seg0_pa1 {
+		pinmux = <STM32_PINMUX('A', 1, AF11)>;
+	};
+
+	/omit-if-no-ref/ lcd_seg1_pa2: lcd_seg1_pa2 {
+		pinmux = <STM32_PINMUX('A', 2, AF11)>;
+	};
+
+	/omit-if-no-ref/ lcd_seg2_pa3: lcd_seg2_pa3 {
+		pinmux = <STM32_PINMUX('A', 3, AF11)>;
+	};
+
+	/omit-if-no-ref/ lcd_seg3_pa6: lcd_seg3_pa6 {
+		pinmux = <STM32_PINMUX('A', 6, AF11)>;
+	};
+
+	/omit-if-no-ref/ lcd_seg4_pa7: lcd_seg4_pa7 {
+		pinmux = <STM32_PINMUX('A', 7, AF11)>;
+	};
+
+	/omit-if-no-ref/ lcd_com0_pa8: lcd_com0_pa8 {
+		pinmux = <STM32_PINMUX('A', 8, AF11)>;
+	};
+
+	/omit-if-no-ref/ lcd_com1_pa9: lcd_com1_pa9 {
+		pinmux = <STM32_PINMUX('A', 9, AF11)>;
+	};
+
+	/omit-if-no-ref/ lcd_com2_pa10: lcd_com2_pa10 {
+		pinmux = <STM32_PINMUX('A', 10, AF11)>;
+	};
+
+	/omit-if-no-ref/ lcd_seg17_pa15: lcd_seg17_pa15 {
+		pinmux = <STM32_PINMUX('A', 15, AF11)>;
+	};
+
+	/omit-if-no-ref/ lcd_seg5_pb0: lcd_seg5_pb0 {
+		pinmux = <STM32_PINMUX('B', 0, AF11)>;
+	};
+
+	/omit-if-no-ref/ lcd_seg6_pb1: lcd_seg6_pb1 {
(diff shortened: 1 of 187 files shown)


dts/st/l4/stm32l476v(c-e-g)tx-pinctrl.dtsi
```diff
@@ -1068,6 +1068,232 @@
 		pinmux = <STM32_PINMUX('E', 6, AF0)>;
 	};
 
+	/* LCD */
+
+	/omit-if-no-ref/ lcd_seg0_pa1: lcd_seg0_pa1 {
```
thegentlecoder, line 1073: @avolmat-st Just to understand, why is this LCD-GPIO pin list different than the one you can find [here](https://github.com/zephyrproject-rtos/zephyr/pull/112647/changes#diff-62418a56758c0ff5aaa5eb3270a610e8cea7b513717b644e25ecca3de3d49e33R52)?

erwango (changes requested):
Please split in 2 commits:
- generation config file update
- new batch of pcintrl.dtsi files

erwango (approved):

dts/st/l4/stm32l476v(c-e-g)tx-pinctrl.dtsi
avolmat-st: I haven't checked all the pins, however, at least this seg0_pa1, seg1_pa2 seems to be correct based on the datasheet document https://www.st.com/resource/en/datasheet/stm32l476rg.pdf.
Cf table Table 16. STM32L476xx pin definitions.

From where are coming the pinctrl settings from your PR ?

avolmat-st: Sorry, seems I didn't see your comment before.

thegentlecoder: Truly that's my fault: I forgot to press "Submit comments", sorry.

I'm referring to page 27 of https://www.st.com/resource/en/user_manual/um1879-discovery-kit-with-stm32l476vg-mcu-stmicroelectronics.pdf, the second manual in https://www.st.com/en/evaluation-tools/32l476gdiscovery.html#documentation 

Here's the image, including also PC3 for VLCD, that is not actually part of the GH08172T itself, but needed to make it work:

Image

avolmat-st: Thanks for the info. From what I understand, so, basically on this board, what is linked to the SEG0 of the LCD display is connected to PA7. This PA7 is also a pin which is controlled by the LCD HW Block so this is fine, however, from the LCD HW Block, this is SEG4.
So, basically, from what I understand, it is just a matter of writting later on the correct value into memory to value to make the right segment turn on.
So my guess is that in the current driver you have, whenever you want to turn on SEG0 of the LCD display, the value given might be actually turning on SEG4 and not SEG0, but since that's how is wired this LCD display, no choice.
Would that make sense ?

thegentlecoder: Maybe yes, I'll check with the board, once I'll have understood how to instantiate a sample application with a shield. BTW, C3 for VLCD is missing:
/omit-if-no-ref/ lcd_vlcd_pc3: lcd_vlcd_pc3 {
		pinmux = <STM32_PINMUX('C', 3, AF11)>;
	};

A final question about this file (just seen now): pin C10, C11 and C12 are used multiple times starting at line 1202, is this intentional?

avolmat-st: I think yes. I mean this is coherent with the datasheet at least. It would be needed to look into more detail about the behavior of the LCD Block to understand the COM / SEG handling.

thegentlecoder: About LCD_VLCD on C3

Image

avolmat-st: Oh, my bad, I didn't put that into the possible naming hence it is missing. Let me fix this then.

thegentlecoder: > Thanks for the info. From what I understand, so, basically on this board, what is linked to the SEG0 of the LCD display is connected to PA7. This PA7 is also a pin which is controlled by the LCD HW Block so this is fine, however, from the LCD HW Block, this is SEG4. So, basically, from what I understand, it is just a matter of writting later on the correct value into memory to value to make the right segment turn on. So my guess is that in the current driver you have, whenever you want to turn on SEG0 of the LCD display, the value given might be actually turning on SEG4 and not SEG0, but since that's how is wired this LCD display, no choice. Would that make sense ?

@avolmat-st , I finally got it, everything is ok, just C3 was missing. Thank you

thegentlecoder: Apparently I'm not allowed to resolve this comment.

avolmat-st: No pb, let me first push the update hal_stm32 PR for the missing C3 and then I can mark this comment as resolved.

PR #391, opened 2026-07-10
stm32Cube: stm32u5 HAL ospi driver adds memory type and size functions

FRASTM:
Add HAL_OSPI_SetMemoryType and HAL_OSPI_SetMemorySize API functions to the stm32U5 HAL driver to set the DCR1 register out of init Same model has the XSPi HAL driver functions.
This is required for the mspi flash ospi driver

Adding to the stm32u5xx/drivers
HAL_OSPI_SetMemoryType(OSPI_HandleTypeDef *hospi, uint32_t Type);
HAL_OSPI_SetDeviceSize(OSPI_HandleTypeDef *hospi, uint32_t Size);

### DIFF
--- a/stm32cube/stm32u5xx/drivers/include/stm32u5xx_hal_ospi.h
+++ b/stm32cube/stm32u5xx/drivers/include/stm32u5xx_hal_ospi.h
@@ -869,6 +869,8 @@ HAL_StatusTypeDef     HAL_OSPI_Abort(OSPI_HandleTypeDef *hospi);
 HAL_StatusTypeDef     HAL_OSPI_Abort_IT(OSPI_HandleTypeDef *hospi);
 HAL_StatusTypeDef     HAL_OSPI_SetFifoThreshold(OSPI_HandleTypeDef *hospi, uint32_t Threshold);
 uint32_t              HAL_OSPI_GetFifoThreshold(const OSPI_HandleTypeDef *hospi);
+HAL_StatusTypeDef     HAL_OSPI_SetMemoryType(OSPI_HandleTypeDef *hospi, uint32_t Type);
+HAL_StatusTypeDef     HAL_OSPI_SetDeviceSize(OSPI_HandleTypeDef *hospi, uint32_t Size);
 HAL_StatusTypeDef     HAL_OSPI_SetTimeout(OSPI_HandleTypeDef *hospi, uint32_t Timeout);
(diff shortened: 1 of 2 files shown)


stm32cube/stm32u5xx/drivers/include/stm32u5xx_hal_ospi.h
```diff
@@ -869,6 +869,8 @@ HAL_StatusTypeDef     HAL_OSPI_Abort(OSPI_HandleTypeDef *hospi);
 HAL_StatusTypeDef     HAL_OSPI_Abort_IT(OSPI_HandleTypeDef *hospi);
 HAL_StatusTypeDef     HAL_OSPI_SetFifoThreshold(OSPI_HandleTypeDef *hospi, uint32_t Threshold);
 uint32_t              HAL_OSPI_GetFifoThreshold(const OSPI_HandleTypeDef *hospi);
+HAL_StatusTypeDef     HAL_OSPI_SetMemoryType(OSPI_HandleTypeDef *hospi, uint32_t Type);
+HAL_StatusTypeDef     HAL_OSPI_SetDeviceSize(OSPI_HandleTypeDef *hospi, uint32_t Size);
```
erwango, line 873: Please update README with internal reference.

erwango (changes requested):
Should it be duplicated to similar HALs (H7, L4, L5, MP2) ?

stm32cube/stm32u5xx/drivers/src/stm32u5xx_hal_ospi.c
```diff
@@ -2550,6 +2550,65 @@ uint32_t HAL_OSPI_GetFifoThreshold(const OSPI_HandleTypeDef *hospi)
...
+/** @brief  Set OSPI Device Size.
+  * @param  hospi : OSPI handle.
+  * @param  Size : Device Size.
+  * @retval HAL status
+  */
+HAL_StatusTypeDef HAL_OSPI_SetDeviceSize(OSPI_HandleTypeDef *hospi, uint32_t Size)
+{
+  HAL_StatusTypeDef status = HAL_OK;
+
+  assert_param(IS_OSPI_MEMORY_SIZE(Size));
```
mathieuchopstm, line 2591: This macro does not exist.

mathieuchopstm (commented):
If those are not called by the HAL, why not have this implementation directly in the corresponding Zephyr driver instead?

PR #393, opened 2026-07-22
stm32cube: stm32h7: fix clang compilation

jciupis:
LLVM defines the __GNUC__ macro but does not support the optimize attribute, at least per the 24.0.0 version:
https://clang.llvm.org/docs/AttributeReference.html

Fix the unknown attribute compilation error by adding a compile-time guard for Clang specifically.


jciupis:
I've just noticed that the file actually comes from https://github.com/STMicroelectronics/cmsis-device-h7 repository, which is a submodule of https://github.com/STMicroelectronics/STM32CubeH7 that is released periodically to this repo.

The change I intended to merge here is already present in the CMSIS repository as https://github.com/STMicroelectronics/cmsis-device-h7/commit/de8243d2c15f87936f28a49fcd9e6f5ba10fc233 and should flow to this repo with a future release.

I'm closing the PR.

PR #399, opened 2026-08-06
scripts: genpinctrl: cleanup & improvements

mathieuchopstm:
Common:
- Unify templates between `genpinctrl` and `genpinctrl_hal2` to simplify maintenance
- Emit final newline in all generated files (README / pinctrl)
- Add missing section in README template creating parasitic changes upon regeneration
- Don't emit unnecessary empty lines in pinctrl files
  - *This is not strictly required but rather a matter of taste; can be discussed*
- Comply with Zephyr Devicetree Style Guidelines by not emitting an empty line right before the dedenting `};` near the end of pinctrl files

`genpinctrl_hal2`:
- use `pathlib.Path` consistently across the script for readability and simplicity
- change script to consume actual STM32Cube HAL2 packages as it claimed to do, even though it actually needed a weirdly organized folder as input

\+ regeneration of all pinctrl files with updated scripts which only creates whitespace changes (no functional difference on generation)


erwango (changes requested):
IIUC pinctrl generation validation action actually depended on the functionality removed by the last HAL2 related commit (running `  python3 scripts/genpinctrl/genpinctrl_hal2.py -p /tmp/STM32Cube_DFP -o dts`).

Note sure why it didn't complain (because there's only one HAL2 based series ? because of poor error checking ?), but it should probably be updated as well.

erwango (approved):

PR #401, opened 2026-08-20
stm32cube: *: delete timebase and MSP templates

mathieuchopstm:
When used in Zephyr, the HAL time base is provided by the System Timer and each peripheral driver enables/configures resources it needs using standard Zephyr APIs like Clock Control or Pinctrl. This makes the HAL Timebase and MSP modules unnecessary, so delete the template files from this repository. Note that for some series, the "template" suffix has been dropped(!) but the files are nevertheless unused.


gautierg-st (approved):

erwango (approved):
Ack'ed.
Usually, I'd have asked a script update. But that would not really be useful if a deep rewrite is pending. Correct ?

mathieuchopstm:
> Ack'ed. Usually, I'd have asked a script update. But that would not really be useful if a deep rewrite is pending. Correct ?

Correct - new script takes this case into account and ignores all `*_template.c` files during import/update.

PR #413, opened 2026-09-22
PSA crypto driver: HASH and AES/CRYP

etienne-lms:
Add a source tree that provides PSA Crypto drivers integration: `stm32_psa_crypto_drivers/`.
The source tree provides the drivers and the pre-generated dispacher files for integration in tf-psa-crypto.

This P-R implements HASH (hash only) and AES/CRYP (transparent ECB/CBC/CTR/CCM/GCM only) based on HAL.

A series of preliminary changes updates some SoCs HAL configuration files to allow the build environment to enable or not HW context backup in related HASH and/or AES/CRYP HAL drivers. Multi-part crypto operations with PSA Crypto driver requires the HW context can be saved/restored between operation steps.


lib/stm32_psa_crypto_drivers/config/stm32_psa_crypto_drivers_config.h
mathieuchopstm, line 1: what about `lib/stm32_psa_crypto_drivers` instead? it seems like a good fit especially since some files have a non-(pure) Apache 2.0 license

lib/stm32_psa_crypto_drivers/include/stm32_soc_hal.h
```diff
@@ -0,0 +1,229 @@
...
+
+#elif /* WL */ \
+      defined(STM32WL55xx) || defined(STM32WL54xx) || \
+      defined(STM32WLE5xx) || defined(STM32WLE4xx) || defined(STM32WL5Mxx)
+#include "stm32wlx.h"
+
+#elif /* WL3 */ \
+      defined(STM32WL33) || defined(STM32WL3XX) || \
+      defined(STM32WL3R) || defined(STM32WL3RX)
+#include "stm32wl3x.h"
```
mathieuchopstm, line 222: Is this a file you defined by hand? because `STM32WL33` is a legacy alias...

stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_hash.c
```diff
@@ -0,0 +1,794 @@
...
+  mbedtls_mutex_unlock(&hash_device.mutex);
+#endif /* MBEDTLS_THREADING_C */
+}
+
+/*
+ * Increment the number of active operations.
+ * Device must be locked when this function is called.
+ * Return true if first operation, false otherwise
+ */
+static bool increment_opration_counter(void)
```
smartin-vossloh: Is the typo (op***e***ration) intended?

Ditto in the decrement API.

And probably other similar typos in other commits.

stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aes_private.h
```diff
@@ -0,0 +1,167 @@
...
+/* Unlock access to the hardware device */
+void stm32_hal_aes_put_hal_hw(void);
+
+/*
+ * The operation context is being destroyed and will no more request access
+ * to the device.
+ */
+void stm32_hal_aes_release_hal_hw(stm32_hal_aes_ctx_t *ctx);
+
+/* Check if the key size is supported and provide the relateed HAL ID */
```
smartin-vossloh: nit: typo: relate***e***d

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aes.c
```diff
@@ -0,0 +1,518 @@
...
+#include <string.h>
+#include <stdlib.h>
+
+/* HAL_AES_DEVICE helper macro to identify the target device for AES crypto */
+#if defined(STM32_HAL_AES_DPA_DRIVER_ENABLED)
+# ifdef STM32_AES_HAL_V2
+#define HAL_AES_DEVICE   HAL_SAES
+# else
+#define HAL_AES_DEVICE   SAES
+# endif /* STM32_AES_HAL_V2 */
```
smartin-vossloh, line 34: @etienne-lms Does the DPA logic mean it will not be possible to use both the AES and SEAS engines at the same time when available?

stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aes.c
```diff
@@ -0,0 +1,518 @@
...
+
+  restore_u32_regs(&ctx->hal_handle.Instance->K0LR, ctx->key, ctx->key_byte_size / 32);
+#else
+  ASSERT(0);
+#endif /* STM32F2 || STM32F4 || STM32F7 || STM32H7 */
+}
+#endif /* STM32_AES_LOCAL_SUSPEND_CONTEXT */
+
+/*
+ * Save STM32 device hardware state in a hash operation context.
```
smartin-vossloh: AES operation context instead of the hash one, right?

Ditto in the restore_context comment.

stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_cipher.c
```diff
@@ -0,0 +1,681 @@
...
+
+cleanup:
+  stm32_cipher_release_context(&ctx);
+
+  return status;
+}
+
+/*
+ * PSA Crypto Driver handler for cipher_decrypt()
+ * For CBC and CTR, IV is locaed at the begining of the input data buffer.
```
smartin-vossloh: typo: loca***t***ed

stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aead.c
```diff
@@ -0,0 +1,1436 @@
...
+                                                          size_t tag_size)
+{
+  switch (ctx->algo_id)
+  {
+  case STM32_HAL_ALG_AES_GCM:
+    if ((tag_size < 4) || (tag_size > 16))
+    {
+      return STM32_HAL_INVALID_ARGUMENT;
+    }
+   break;
```
smartin-vossloh: nit: weird indentation

Ditto below, and duplicated `break` statement.

stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aead.c
```diff
@@ -0,0 +1,1436 @@
...
+
+  /* First authentication block in little-endian format */
+  swap_u32_endianness(ctx->iv, iv_data, sizeof(iv_data));
+
+  return STM32_HAL_SUCCESS;
+}
+
+/*
+ * Prepare CCM B1 data buffer (headered buffer for additional data).
+ * This requirs the number of additional data is known.
```
smartin-vossloh: typo: requir***e***s

stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aead.c
```diff
@@ -0,0 +1,1436 @@
...
+#ifdef PRINT_ENTRY
+  printf("%s() line %u (sizes: nonce %zu)\n", __func__, __LINE__, nonce_length);
+#endif
+
+  switch (ctx->algo_id)
+  {
+  case STM32_HAL_ALG_AES_GCM:
+    if (nonce_length != 12)
+    {
+printf("\n--\n-- GCM nonce != 12byte => not supported\n--\n");
```
smartin-vossloh: Could probably be removed, no?

stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aead.c
```diff
@@ -0,0 +1,1436 @@
...
+
+  return STM32_HAL_SUCCESS;
+}
+
+STM32_HalStatusTypeDef STM32_HalAeadSetNonce(stm32_hal_aes_ctx_t *ctx,
+                                             const uint8_t *nonce, size_t nonce_length)
+{
+#ifdef PRINT_ENTRY
+  printf("%s() line %u (sizes: nonce %zu)\n", __func__, __LINE__, nonce_length);
+#endif
```
smartin-vossloh: Should this debug traces be kept?

Ditto in below functions.

stm32cube/stm32wlxx/drivers/include/stm32wlxx_hal_conf.h
```diff
@@ -182,7 +182,9 @@ extern "C" {
```
smartin-vossloh, line 182: @etienne-lms Duplicated message in the commit log.

smartin-vossloh (commented):
Thanks a lot for this work.

lib/stm32_psa_crypto_drivers/include/stm32_soc_hal.h
etienne-lms: Yes, I wrote this file manually. I don't think there is any equivalent in the STM32Cube* distributions.

lib/stm32_psa_crypto_drivers/config/stm32_psa_crypto_drivers_config.h
etienne-lms: I'm fine moving to your proposed path.

stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_hash.c
etienne-lms: Likely many similar typos. Thank to pointing that. I'll fix.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aes.c
etienne-lms: Currently yes, you can use only one of them. The implementation could evolve to provide support for both and select the HW peripheral upon... a specific key attribute?. Note SAES will be use for opaque keys operations (not implemented yet) but it will be able to execute opaque keys as transparent keys operations.

stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aes.c
etienne-lms: Thanks, copy/paste mistake of mine.

stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aead.c
etienne-lms: Indeed, I missed it!!!

stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aead.c
etienne-lms: I'll create some kind of debug trace macros to implement these. It's somewhat useful when porting to new SoCs.

stm32cube/stm32wlxx/drivers/include/stm32wlxx_hal_conf.h
etienne-lms: Thanks.
Thank a lot for your feedback. I'll update the series.

etienne-lms:
@smartin-vossloh's comments addressed. [Diff here](https://github.com/zephyrproject-rtos/hal_stm32/compare/840b42c0f0e93a93268b575bf302545e75d00f9b..359ce9180dbc8d20bb374d0ac10bd3fcdd56b0d5).

I'll change the location path later.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aes.c
smartin-vossloh: > Note SAES will be use for opaque keys operations (not implemented yet) but it will be able to execute opaque keys as transparent keys operations.

Yeap, and I am definitely interested by this feature.
I have not dived much in the PSA specs, but I expect this use-case would be supported by the API.
But I don't think I will have time to spend on the opaque key support using PSA API soon. :confused: 

> The implementation could evolve to provide support for both and select the HW peripheral upon... a specific key attribute?

Or upon some engine capabilities and preferences, i.e. automatically reduce the list to those supporting the algo, then if still several options, use the preference (e.g.: security over rapidity, or the opposite)...

Just sharing wild ideas... :slightly_smiling_face:

etienne-lms:
Moved to `/lib/stm32_psa_crypto_drivers/` and fixed a missing `#endif` in a intermediate commit.
[Diff here](https://github.com/zephyrproject-rtos/hal_stm32/compare/359ce9180dbc8d20bb374d0ac10bd3fcdd56b0d5..e6c3972740381e04074840f3edff94015a9ab86c).

lib/stm32_psa_crypto_drivers/LICENSE.txt
```diff
@@ -0,0 +1,6 @@
+This software component is provided to you as part of a software package and
+applicable license terms are in the  Package_license file. If you received this
```
erwango, line 2: It seems we're referencing `the  Package_license file` without defining it.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/LICENSE.md
erwango, line 1: Do we actually need this file, since we have lib/stm32_psa_crypto_drivers/LICENSE.md ?
Ditto for LICENCE.txt

erwango (changes requested):
Please update https://github.com/zephyrproject-rtos/hal_stm32/blob/main/REUSE.toml while you're at.

etienne-lms:
Update to use `SOC_FAMILY_` prefixed config directives.

erwango (changes requested):
Please add a README similar to https://github.com/zephyrproject-rtos/hal_stm32/edit/main/lib/vc8000nanoe/README.rst

For the part that are unknown, make it clear that this will be filled as soon as available

etienne-lms:
I've added a README file and fixed an issue with STM32WBxx product line (where AES instance(s) have specific naming and related macro names). [Diff here](https://github.com/zephyrproject-rtos/hal_stm32/compare/f8b33ec2bbcbc231c3aad7813013f8c46af47ae3..58fdcd195a07be114390046611c41df07d005977).

etienne-lms:
Rebased to leverage STM32C5xx support recently merged and fixed a typo affecting STM32WLxx product lines [replaced `stm32wlx.h` with `stm32wlxx.h`](https://github.com/zephyrproject-rtos/hal_stm32/compare/58fdcd195a07be114390046611c41df07d005977..ad244b3b756ba477834c077049435a65a196fa20#diff-611709273fd56a5f2a8a1d7dd3dcc4b338a7d1ee4c1443ce48d4d9e113054259).

lib/stm32_psa_crypto_drivers/include/stm32_soc_hal.h
mathieuchopstm: I would drop the legacy aliases in this case. Non blocking.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/include/stm32_hal_core_config.h
```diff
@@ -0,0 +1,98 @@
...
+  * Enable STM32 DPA resistance core for SAES, secure crypto operations.
+  * When this option is enabled, AES-ECB/CBC modes become DPA-protected.
+  *
+  * Uncomment the macro to enable DPA
+  *
+  * Requires STM32_HAL_AES_CIPHER_DRIVER_ENABLED or STM32_HAL_AES_AEAD_DRIVER_ENABLED.
+  *
+  */
+
+//#define STM32_HAL_AES_DPA_DRIVER_ENABLED
```
mathieuchopstm, line 79: Why commented out by default?

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/include/stm32_hal_hash.h
```diff
@@ -0,0 +1,127 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* Includes ------------------------------------------------------------------*/
+#include "stm32_hal_types.h"
+#include "stm32_hal_core_config.h"
+
+#define STM32_HASH_BLOCK_MAX_SIZE  ((size_t) 128)    /*!< HW handles 512 bits, ie 64 bytes */
```
mathieuchopstm, line 30: Not clear how 512 bits/64 bytes becomes 128...

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/include/stm32_hal_hash.h
```diff
@@ -0,0 +1,127 @@
...
+#define STM32_HASH_EXTRA_BYTES ((size_t)   4)        /*!< One supplementary word on first block */
+#define STM32_HASH_TIMEOUT     ((uint32_t) 3)
```
mathieuchopstm, line 32: What is the unit?

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/include/stm32_hal_hash.h
```diff
@@ -0,0 +1,127 @@
...
+ * STM32_HASH_HAL_V* helper macros to differenciate between the several HAL
+ * API versions.
+ */
+#if defined(HAL_VERSION_MAIN) && (HAL_VERSION_MAIN == 2)
+#define STM32_HASH_HAL_V2
+#elif defined(HAL_HASH_VERSION) && (HAL_HASH_VERSION == 200)
+#define STM32_HASH_HAL_V1_2
+#else
+#define STM32_HASH_HAL_V1_0
+#endif
```
mathieuchopstm, line 44: It would be more readable IMO to have:
* `STM32_HASH_HAL2`: for HAL2 format
* `STM32_HASH_HAL1_V2`: for HAL1 format with `HASH_VERSION 200`
* `STM32_HASH_HAL1_V1`: for HAL1 format w/o `HASH_VERSION`

Similar comment for the AES header, although it's less of an issue there because the special HAL1 version is identified by series name.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aead.c
```diff
@@ -0,0 +1,1421 @@
...
+    return STM32_HAL_ERROR_NOT_SUPPORTED;
+  }
+
+  ctx->aead_config.tag_size = tag_size;
+
+  return STM32_HAL_SUCCESS;
+}
+
+/*
+ * Prepare CCM B0 data (stored in SoC IVR registers
```
mathieuchopstm, line 119: missing closing paren

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aead.c
```diff
@@ -0,0 +1,1421 @@
...
+STM32_HalStatusTypeDef STM32_HalAeadSetLengths(stm32_hal_aes_ctx_t *ctx,
+                                               size_t ad_length,
+                                               size_t plaintext_length)
+{
+  LOG_ENTRY("%s() line %u (aad %zu, msg %zu)\n", __func__, __LINE__,
+            ad_length, plaintext_length);
+
+  /*
+   * Nonce/IV shall not be set yet.
+   * Can't change AAD size once additional data are received .
```
mathieuchopstm, line 753:
```suggestion
   * Can't change AAD size once additional data are received.
```

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aead.c
```diff
@@ -0,0 +1,1421 @@
...
+  if ((ctx->aead_config.current_aad_size + input_length) < input_length)
+  {
+    return STM32_HAL_INVALID_ARGUMENT;
+  }
+
+  switch (ctx->algo_id)
+  {
+  case STM32_HAL_ALG_AES_GCM:
+    /*
+     * GCM allows to  not have the AAD size pre-defined with psa_aead_set_lengths().
```
mathieuchopstm, line 837:
```suggestion
     * GCM allows to not have the AAD size pre-defined with psa_aead_set_lengths().
```

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aes_private.h
```diff
@@ -0,0 +1,170 @@
...
+ * 32-bit integer manipulation macros (big endian)
+ */
+#ifndef GET_UINT32_BE
+#define GET_UINT32_BE(n,b,i)                 \
+  do {                                       \
+    (n) = ((uint32_t)(b)[(i)    ] << 24 ) |  \
+          ((uint32_t)(b)[(i) + 1] << 16 ) |  \
+          ((uint32_t)(b)[(i) + 2] <<  8 ) |  \
+          ((uint32_t)(b)[(i) + 3]       );   \
+  } while(0)
```
mathieuchopstm, line 65: A macro that evaluates to an rvalue (like `SWAP_UINT32_ENDIANNESS`) would be more readable imo.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aes_private.h
```diff
@@ -0,0 +1,170 @@
...
+                                       size_t byte_size)
+{
+  size_t i;
+
+  for (i = 0; i < (byte_size / sizeof(uint32_t)); i++)
+  {
+    /* TODO; use Arm REV assembly instruction */
+    GET_UINT32_BE(out[i], in, 4U * i);
+  }
+}
```
mathieuchopstm, line 94: `read_u32_swapped()` would make more sense as a name given the definition of `SWAP_UINT32_ENDIANNESS` above.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_rng.c
```diff
@@ -0,0 +1,198 @@
...
+{
+  COMPILER_BARRIER();
+  uint8_t newValue;
+  do
+  {
+    newValue = __LDREXB(valuePtr) + delta;
+  } while (__STREXB(newValue, valuePtr));
+  COMPILER_BARRIER();
+  return newValue;
+}
```
mathieuchopstm, line 39: ARMv6-M does not have exclusive load/store instructions. TBC if this is a possible target of this implementation.

lib/stm32_psa_crypto_drivers/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+
+zephyr_sources(stm32_psa_transparent_drivers/interface/src/stm32_hal_psa_translator.c)
+
+if(CONFIG_SOC_FAMILY_STM32_PSA_CRYPTO_DRIVERS_AES OR
+   CONFIG_SOC_FAMILY_STM32_PSA_CRYPTO_DRIVERS_CRYP OR
+   CONFIG_SOC_FAMILY_STM32_PSA_CRYPTO_DRIVERS_SAES)
+  zephyr_compile_definitions(USE_STM32_HAL_PSA_AES_DRIVER)
+  zephyr_compile_definitions_ifdef(CONFIG_SOC_FAMILY_STM32_PSA_CRYPTO_DRIVERS_SAES,
+    USE_STM32_HAL_PSA_AES_DPA_DRIVER=1
+  )
```
mathieuchopstm, line 29: `CONFIG_STM32CUBE_xxx` options declared in `modules/Kconfig.stm32` seem like they would be more appropriate.

README.rst
```diff
@@ -155,5 +155,6 @@ For some specific features, in order to benefit from already available and
 validated code. Libraries from STM32Cube packages are extracted and made
 available in this module under `lib` folder.
 
-These libraries are updated automatically by the ``ble_library.py`` script in a separate commit
-directly after their STM32Cube package.
+These libraries (but `lib/stm32_psa_crypto_drivers`) are updated automatically
```
mathieuchopstm, line 158:
```suggestion
These libraries (except `lib/stm32_psa_crypto_drivers`) are updated automatically
```

The vcnano8000e library is not updated either by the way. This should probably reworded to:
```
The following libraries are updated automatically by the ``ble_library.py`` script
in a separate commit directly after their STM32Cube package:
* ``stm32wb0``
* ``stm32wba``
* ...

The following libraries are updated manually when appropriate or necessary:
* ``vcnano8000e``
* ``stm32_psa_crypto_drivers``
```

mathieuchopstm (commented):
Didn't review the drivers very thoroughly, mostly comments from things I spotted while scrolling.

None of the comments are really blocking except for the one in `lib/README.rst` which I'd really like to be addressed.

lib/stm32_psa_crypto_drivers/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
+# SPDX-License-Identifier: Apache-2.0
+
+zephyr_library_link_libraries(tfpsacrypto)
```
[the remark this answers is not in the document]
mathieuchopstm: I believe we currently share the `zephyr_library()` from `stm32cube/CMakeLists.txt`. This is quite ugly, but fixing this is out of scope.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aead.c
```diff
@@ -0,0 +1,1421 @@
...
+  {
+    b1_header_size = 6;
+  }
+
+  /*
+   * Extra bytes to deal with data padding such that
+   * the resulting string can be partitioned into words
+   */
+  b1_padding = (ad_length + b1_header_size) % 4U;
+  b1_length = ad_length + b1_header_size + b1_padding;
```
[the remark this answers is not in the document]
mathieuchopstm: Proposed fix is not completely correct since the length does need to be increased, I believe.

kshypachov:
Hi @etienne-lms, @erwango,

Thanks for publishing the STM32 PSA Crypto drivers. Is PKA support (P-256 ECDSA, ECDH, key generation) planned for
STM32U5 as well?

etienne-lms:
Yes, PKA for EC is on its way. I don't think it will be part of the next release tag.
I hope the proposed hash and AES integration will be merged before v4.5.0 release tag. Any review contribution to this P-R and https://github.com/zephyrproject-rtos/zephyr/pull/119888 will help.

lib/stm32_psa_crypto_drivers/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+# Crypto drivers need access to internal TF-PSA-Crypto headers
+zephyr_library_include_directories($<TARGET_PROPERTY:tfpsacrypto,INCLUDE_DIRECTORIES>)
```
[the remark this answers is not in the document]
etienne-lms: It is intended that no CMake file here adds `dispatch/psa_crypto_driver_wrappers_no_static.c`. This contribution is intended to be used with a Zephyr configuration that sets `CONFIG_TF_PSA_CRYPTO_DISPATCH_DIR` to the path of the `dispatch/` directory provided here. See `TF_PSA_CRYPTO_DISPATCH_DIR` default value set in a `Kconfig` file in [P-R #119888](https://github.com/zephyrproject-rtos/zephyr/pull/119888/changes).

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aead.c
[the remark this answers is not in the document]
etienne-lms: Using a round helper macro/function would make this clearer. 
I'll go for
```c
  /* Round up buffer length to 32bit words */
  b1_length = ((ad_length + b1_header_size + 3U) / 4U) * 4U;

File: lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aead.c [hunk 28]
```diff
@@ -0,0 +1,1421 @@
...
+    b1_padded_addr[1] = (uint8_t)(ad_length & 0xFF);
+  }
+  else
+  {
+    b1_padded_addr[0] = 0xFF;
+    b1_padded_addr[1] = 0xFE;
+    b1_padded_addr[2] = (uint8_t)(ad_length & 0xFF);
+    b1_padded_addr[3] = (uint8_t)((ad_length >> 8) & 0xFF);
+    b1_padded_addr[4] = (uint8_t)((ad_length >> 16) & 0xFF);
+    b1_padded_addr[5] = (uint8_t)((ad_length >> 24) & 0xFF);
```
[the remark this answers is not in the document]
etienne-lms: I think that is right. Actually, I see that tf-psa-crypto implementation, at least up to v1.2.0 rejects CCM AAD length above 0xff00, hence this code is never used. I'll remove it and will restore a tested implementation later.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aead.c
```diff
@@ -0,0 +1,1421 @@
...
+                                                        size_t length)
+{
+  stm32_hal_status_t hal_status = HAL_OK;
+  STM32_HalStatusTypeDef status;
+
+  status = stm32_hal_aes_get_hal_hw(ctx, stm32_aead_hal_configure);
+  if (status != STM32_HAL_SUCCESS)
+  {
+    goto exit;
+  }
```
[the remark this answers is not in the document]
etienne-lms: That's right.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aead.c
```diff
@@ -0,0 +1,1421 @@
...
+
+  /* If no data, ensure the HAL is informed (process O bytes) */
+  if (ctx->aead_config.current_message_size == 0)
+  {
+    if (stm32_aead_process_buffer(ctx, NULL, NULL, 0) != STM32_HAL_SUCCESS)
```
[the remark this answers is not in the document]
etienne-lms: Right, i'll fix:
```patch
-  status = stm32_hal_aes_get_hal_hw(ctx, stm32_aead_hal_configure);
-  if (status != STM32_HAL_SUCCESS)
-  {
-    return status;
-  }
-
   /* If no data, ensure the HAL is informed (process O bytes) */
   if (ctx->aead_config.current_message_size == 0)
   {
     if (stm32_aead_process_buffer(ctx, NULL, NULL, 0) != STM32_HAL_SUCCESS)
     {
-       status = STM32_HAL_ERROR;
-       goto exit;
+      return STM32_HAL_ERROR;
     }
   }
 
+  status = stm32_hal_aes_get_hal_hw(ctx, stm32_aead_hal_configure);
+  if (status != STM32_HAL_SUCCESS)
+  {
+    return status;
+  }
+
```

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aead.c
```diff
@@ -0,0 +1,1421 @@
...
+  size_t proc_size;
+
+  LOG_ENTRY("%s() line %u (in %zu, out-buf %zu)\n", __func__, __LINE__, input_length, output_size);
+
+  if (input_length != 0)
+  {
+    if (output_size < input_length)
+    {
+      return STM32_HAL_BUFFER_TOO_SMALL;
+    }
```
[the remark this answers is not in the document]
etienne-lms: Right, i'll check that.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aes.c
```diff
@@ -0,0 +1,530 @@
...
+  * If no LICENSE file comes with this software, it is provided AS-IS.
+  *
+  ******************************************************************************
+  */
+
+/* Includes ------------------------------------------------------------------*/
+
+#include "stm32_hal_aes.h"
+#include "stm32_hal_aes_private.h"
+#include "stm32_hal_private.h"
```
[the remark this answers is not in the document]
etienne-lms: Thanks

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aes.c
```diff
@@ -0,0 +1,530 @@
...
+/* HAL_ENABLE_CLOCK() and HAL_DISABLE_CLOCK() helper macros  */
+#ifdef STM32_HAL_AES_DPA_DRIVER_ENABLED
+ #ifdef STM32_AES_HAL_V2
+#define HAL_ENABLE_CLOCK() \
+  do { HAL_RCC_RNG_EnableClock(); HAL_RCC_SAES_EnableClock(); } while (0)
+#define HAL_DISABLE_CLOCK() \
+  do { HAL_RCC_SAES_DisableClock(); /* Leave RNG On */ } while (0)
+ #else
+#define HAL_ENABLE_CLOCK()      __HAL_RCC_SAES_CLK_ENABLE
+#define HAL_DISABLE_CLOCK()     __HAL_RCC_SAES_CLK_DISABLE
```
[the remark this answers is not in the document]
etienne-lms: Indeed!

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aes.c
```diff
@@ -0,0 +1,530 @@
...
+  if (ctx->algo_id == STM32_HAL_ALG_AES_CCM)
+  {
+    restore_u32_regs(&ctx->hal_handle.Instance->CSGCM0R, ctx->hw_ctx.suspend, 8);
+  }
+  if (ctx->algo_id == STM32_HAL_ALG_AES_GCM)
+  {
+    restore_u32_regs(&ctx->hal_handle.Instance->CSGCMCCM0R, ctx->hw_ctx.suspend, 8);
+  }
+
+  restore_u32_regs(&ctx->hal_handle.Instance->K0LR, ctx->key, ctx->key_byte_size / 32);
```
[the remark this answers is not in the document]
etienne-lms: Right. I need to restore a test for this. I should have seen failures.
I'll fix. Thanks.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aes.c
```diff
@@ -0,0 +1,530 @@
...
+  HAL_AES_RestoreContext(&ctx->hal_handle, &ctx->hw_ctx);
+#elif defined(STM32_AES_HAL_V1_L4)
+  uint32_t key_size_id;
+  STM32_HalStatusTypeDef status;
+
+  status = stm32_hal_aes_key_size_id(ctx->key_byte_size, &key_size_id);
+  ASSERT(status == STM32_HAL_SUCCESS);
+  HAL_CRYPEx_Write_KeyRegisters(&ctx->hal_handle, (uint8_t *)ctx->key, key_size_id);
+
+  HAL_CRYPEx_Write_IVRegisters(&ctx->hal_handle, (uint8_t *)&ctx->hw_ctx.cr);
```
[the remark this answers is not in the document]
etienne-lms: True

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aes.c
```diff
@@ -0,0 +1,530 @@
...
+  memset(&ctx->hal_handle, 0, sizeof(ctx->hal_handle));
+
+#ifdef STM32_AES_HAL_V2
+  if (HAL_AES_Init(&ctx->hal_handle, HAL_AES_DEVICE) != HAL_OK)
+  {
+    return STM32_HAL_ERROR;
+  }
+#else /* STM32_AES_HAL_V2 */
+# if defined(STM32_HAL_AES_DPA_DRIVER_ENABLED)
+  ctx->hal_handle_aes.Instance = HAL_AES_DEVICE;
```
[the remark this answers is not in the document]
etienne-lms: +1

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_cipher.c
```diff
@@ -0,0 +1,681 @@
...
+    case STM32_HAL_ALG_AES_CTR:
+      break;
+    default:
+      return STM32_HAL_INVALID_ARGUMENT;
+  }
+
+  if (iv_length > sizeof(ctx->iv))
+  {
+    return STM32_HAL_INVALID_ARGUMENT;
+  }
```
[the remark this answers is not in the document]
etienne-lms: +1

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_cipher.c
```diff
@@ -0,0 +1,681 @@
...
+  size_t remaining_input_size = input_length;
+  size_t updated_output_size = 0;
+  size_t proc_size;
+  STM32_HalStatusTypeDef status;
+
+  /* On non-CTR, Output buffer should at least the size of both the pending and new input data */
+  if ((ctx->algo_id != STM32_HAL_ALG_AES_CTR) && (output_size < (input_length + ctx->sbuf_len)))
+  {
+    return STM32_HAL_BUFFER_TOO_SMALL;
+  }
```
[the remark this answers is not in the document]
etienne-lms: Right.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_cipher.c
```diff
@@ -0,0 +1,681 @@
...
+    status = STM32_HalCipherSetIV(&ctx, p_iv, iv_length);
+    if (status != STM32_HAL_SUCCESS)
+    {
+      goto cleanup;
+    }
+  }
+
+  status = STM32_HalCipherUpdate(&ctx, p_plaintext, plaintext_length,
+                                 p_ciphertext, plaintext_length,
+                                 p_ciphertext_length);
```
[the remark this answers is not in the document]
etienne-lms: Thanks.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_rng.c
```diff
@@ -0,0 +1,198 @@
...
+  *
+  * This software is licensed under terms that can be found in the LICENSE file
+  * in the root directory of this software component.
+  * If no LICENSE file comes with this software, it is provided AS-IS.
+  *
+  ******************************************************************************
+  */
+
+/* Includes ------------------------------------------------------------------*/
+#include "stm32_hal_rng.h"
```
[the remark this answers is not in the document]
etienne-lms: True. Actually this file is never built in the scope of the Zephyr application. Since it's still part of the STM32 HAL PSA cryupto drivers, I prefer to leave the source code here. I'll add the header file that was missing.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_rng.c
```diff
@@ -0,0 +1,198 @@
...
+  RNG_ConfigTypeDef rng_conf;
+
+  /*  We're only supporting a single user of RNG */
+  if (atomic_incr_u8(&users, 1U) > 1U)
+  {
+    return status;
+  }
+
+  /* Select RNG clock source */
+  __HAL_RCC_RNG_CONFIG(RNGCLKSOURCE_HSI);
```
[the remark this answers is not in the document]
etienne-lms: The macro comes from the header file that was missing, see the review comment right above.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_rng.c
```diff
@@ -0,0 +1,198 @@
...
+    {
+      return PSA_ERROR_HARDWARE_FAILURE;
+    }
+    init_ctx = 1U;
+  }
+
+  if (RNG_GetBytes(output, output_size, output_length) != 0)
+  {
+    return PSA_SUCCESS;
+  }
```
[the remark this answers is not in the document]
etienne-lms: Thanks. Indeed I need to update this source file, even if Zephyr will not use it.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/config/stm32_hal_psa_transparent_driver_config_template.h
```diff
@@ -0,0 +1,145 @@
...
+  *        AES-ECB/CBC modes become DPA-protected.
+  *
+  * @note Using DPA resistance degrades the performance.
+  *
+  * Uncomment the macro to enable DPA
+  *
+  */
+#if defined(USE_STM32_HAL_PSA_AES_DPA_DRIVER)
+#define STM32_HAL_AES_DPA_PSA_DRIVER_ENABLED
+#endif
```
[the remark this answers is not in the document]
etienne-lms: I'll update.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_cipher.c
```diff
@@ -0,0 +1,681 @@
...
+    status = STM32_HAL_SUCCESS;
+    goto cleanup;
+  }
+
+  /* Supported ECB, CBC and CTR expect block aligned intpu data */
+  if ((plaintext_length % STM32_CIPHER_BLOCK_SIZE) != 0)
+  {
+    status = STM32_HAL_INVALID_ARGUMENT;
+    goto cleanup;
+  }
```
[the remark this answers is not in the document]
etienne-lms: Thanks

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_cipher.c
```diff
@@ -0,0 +1,681 @@
...
+  {
+    return status;
+  }
+
+  /* Supported ECB, CBC and CTR expect block aligned intpu data */
+  if ((ciphertext_length % STM32_CIPHER_BLOCK_SIZE) != 0)
+  {
+    status = STM32_HAL_INVALID_ARGUMENT;
+    goto cleanup;
+  }
```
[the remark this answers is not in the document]
etienne-lms: Thanks

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_rng.c
```diff
@@ -0,0 +1,198 @@
...
+{
+  psa_status_t status = PSA_ERROR_HARDWARE_FAILURE;
+  uint32_t dummy;
+  RNG_ConfigTypeDef rng_conf;
+
+  /*  We're only supporting a single user of RNG */
+  if (atomic_incr_u8(&users, 1U) > 1U)
+  {
+    return status;
+  }
```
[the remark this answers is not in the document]
etienne-lms: +1

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/interface/src/stm32_hal_psa_aead.c
```diff
@@ -0,0 +1,170 @@
...
+                                          STM32_HalAesAlgTypeDef *hal_algo,
+                                          uint8_t *tag_length)
+{
+  if (PSA_ALG_IS_AEAD(alg))
+  {
+    *tag_length = PSA_ALG_AEAD_GET_TAG_LENGTH(alg);
+
+    switch (PSA_ALG_AEAD_WITH_DEFAULT_LENGTH_TAG(alg))
+    {
+#if defined(CONFIG_PSA_WANT_ALG_GCM)
```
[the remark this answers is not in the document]
etienne-lms: My mistake! Thanks.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/interface/src/stm32_hal_psa_hash.c
```diff
@@ -0,0 +1,186 @@
...
+    case PSA_ALG_MD5:
+      *hal_algo = STM32_HAL_ALG_MD5;
+      break;
+#endif /* PSA_WANT_ALG_SHA_1 */
+
+#if defined(PSA_WANT_ALG_SHA1)
+    case PSA_ALG_SHA1:
+      *hal_algo = STM32_HAL_ALG_SHA1;
+      break;
+#endif /* PSA_WANT_ALG_SHA_1 */
```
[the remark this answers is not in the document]
etienne-lms: Indeed

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/include/stm32_hal_hash.h
etienne-lms: typo: SHA384/SHA512x use a 1024 bit internal blocks for porcessing data. I'll fix. Thanks.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/include/stm32_hal_hash.h
etienne-lms: +1

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aes_private.h
etienne-lms: Right, would be better. I'll update.

lib/stm32_psa_crypto_drivers/README.rst
```diff
@@ -0,0 +1,26 @@
...
+Purpose:
+  This library interfaces TF PSA Crypto Drivers framework with the STM32
+  HAL drivers for the various STM32 SoCs and product lines.
+
+Description:
+  This library provides TF PSA Crypto drivers dispatcher, configuration
+  and drivers.
+
+Maintained-by:
+  Internal
```
erwango, line 20:
```suggestion
  External
```
Since it's done by ST

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_rng.c
etienne-lms: Actually we don't build this file in Zephyr. It's there because it's part of the driver that will be delivered in Cube release (or like).

erwango (changes requested):

README.rst
etienne-lms: Nice suggestion. Thanks.

lib/stm32_psa_crypto_drivers/README.rst
etienne-lms: Ok, thanks.

etienne-lms:
Review comments addressed (expect https://github.com/zephyrproject-rtos/hal_stm32/pull/413#discussion_r4083860341 that I missed).

lib/stm32_psa_crypto_drivers/CMakeLists.txt
etienne-lms: I have used `CONFIG_USE_STM32_HAL_PSA_CRYPTO_xxx` to mimic other STM32 HAL related config symbols.

lib/stm32_psa_crypto_drivers/include/stm32_soc_hal.h
```diff
@@ -0,0 +1,229 @@
...
+      defined(STM32MP251Fxx) || defined(STM32MP251Dxx) || defined(STM32MP251Cxx) || defined(STM32MP251Axx) || \
+      /* MP21 */ \
+      defined(STM32MP215Fxx) || defined(STM32MP215Dxx) || defined(STM32MP215Cxx) || defined(STM32MP215Axx) || \
+      defined(STM32MP213Fxx) || defined(STM32MP213Dxx) || defined(STM32MP213Cxx) || defined(STM32MP213Axx) || \
+      defined(STM32MP211Fxx) || defined(STM32MP211Dxx) || defined(STM32MP211Cxx) || defined(STM32MP211Axx) || \
+      /* MP21 */ \
+      defined(STM32MP235Fxx) || defined(STM32MP235Dxx) || defined(STM32MP235Cxx) || defined(STM32MP235Axx) || \
+      defined(STM32MP233Fxx) || defined(STM32MP233Dxx) || defined(STM32MP233Cxx) || defined(STM32MP233Axx) || \
+      defined(STM32MP231Fxx) || defined(STM32MP231Dxx) || defined(STM32MP231Cxx) || defined(STM32MP231Axx)
+#include "stm32mp2xx.h"
```
erwango: mp13 is missing, but probably we don't need mpx at all ?

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/config/stm32_hal_psa_transparent_driver_config_template.h
```diff
@@ -0,0 +1,133 @@
...
+/* Define to prevent recursive inclusion -------------------------------------*/
+#ifndef STM32_HAL_PSA_TRANSPARENT_DRIVER_CONFIG_H
+#define STM32_HAL_PSA_TRANSPARENT_DRIVER_CONFIG_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+
+#if defined(USE_STM32_HAL_PSA_RNG_DRIVER)
```
erwango: If that the correct define here ?

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/config/stm32_hal_psa_transparent_driver_config_template.h
```diff
@@ -0,0 +1,133 @@
...
+  * Uncomment a macro to enable the STM32 HAL PSA RNG module for STM32 hardware
+  * RNG crypto accelerator.
+  *
+  * Requires STM32_HAL_PSA_DRIVER_ENABLED.
+  *
+  * This configuration is disabled in Zephyr since we're using Zephyr native
+  * entropy driver, see CONFIG_MBEDTLS_PSA_DRIVER_GET_ENTROPY and
+  * CONFIG_MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG.
+  */
+#if defined(USE_STM32_HAL_PSA_RNG_DRIVER)
```
erwango: We're already under `#if defined(USE_STM32_HAL_PSA_RNG_DRIVER)` from line 29, but maybe the error is there and not here

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/include/stm32_hal_core_config.h
```diff
@@ -0,0 +1,103 @@
...
+extern "C" {
+#endif
+
+/**
+  * \def STM32_HAL_HASH_DRIVER_ENABLED
+  *
+  * Enables STM32 HAL hash core to use STM32 hardware HASH crypto
+  * accelerator that manages hash operations.
+  *
+  * Uncomment a macro to enable the STM32 HAL hash core.
```
erwango: Macro is not commented. Same on each marcro

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_rng.c
```diff
@@ -0,0 +1,206 @@
...
+    if (HAL_RNGEx_RecoverSeedError(&handle) != HAL_OK)
+    {
+      goto exit;
+    }
+  }
+
+  status = PSA_SUCCESS;
+
+exit:
+  if (status != STM32_HAL_SUCCESS)
```
erwango: PSA_SUCCESS ?

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/include/stm32_hal_hash.h
```diff
@@ -0,0 +1,128 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* Includes ------------------------------------------------------------------*/
+#include "stm32_soc_hal.h"
+#include "stm32_hal_core_config.h"
+#include "stm32_hal_types.h"
+
+#define STM32_HASH_BLOCK_MAX_SIZE  ((size_t) 128)    /*!< Max block size to support SHA384/SHA512 */
```
mathieuchopstm:
```suggestion
#define STM32_HASH_BLOCK_MAX_SIZE  ((size_t) 128)    /*!< Max block size in bytes (128 bytes = 1024-bit for SHA384/SHA512 support) */
```

at least include the unit (bytes)

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aes_private.h
```diff
@@ -0,0 +1,171 @@
...
+  } while(0)
+#endif /* !SWAP_UINT32_ENDIANNESS */
+
+#ifndef SWAP_UINT32_ENDIANESS
+#define SWAP_UINT32_ENDIANESS(val)        \
+  ((((uint32_t)(val) & 0xff) << 24) |     \
+   (((uint32_t)(val) & 0xff00) << 8) |    \
+   (((uint32_t)(val) >> 8) & 0xff00) |    \
+   (((uint32_t)(val) >> 24) & 0xff))
+#endif /* !SWAP_UINT32_ENDIANESS */
```
mathieuchopstm: cont of https://github.com/zephyrproject-rtos/hal_stm32/pull/413#discussion_r4083921103: now you have two macros with the same name???

What I was suggesting was:
```c
// renmaed args for clarity
#define READ_U32_BE(ptr, off) \
    (((uint32_t)((ptr)[(off)    ]) << 24) |
     ((uint32_t)((ptr)[(off) + 1]) << 16) |
     ((uint32_t)((ptr)[(off) + 2]) <<  8) |
     ((uint32_t)((ptr)[(off) + 3])      ))

/* usage */
n = READ_U32_BE(ptr, off);
```

lib/stm32_psa_crypto_drivers/README.rst
```diff
@@ -0,0 +1,26 @@
+STM32 PSA Crypto Drivers library
+################################
+
+Origin:
+  ST Microelectronics
```
mathieuchopstm:
```suggestion
  STMicroelectronics
```

README.rst
```diff
@@ -155,5 +155,12 @@ For some specific features, in order to benefit from already available and
 validated code. Libraries from STM32Cube packages are extracted and made
 available in this module under `lib` folder.
```
mathieuchopstm: off-topic: this should be
```suggestion
available in this module under ``lib`` folder.
```

lib/stm32_psa_crypto_drivers/CMakeLists.txt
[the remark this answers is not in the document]
mathieuchopstm: Maybe a comment in the CMakeLists to indicate why this file is needed could be interesting.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aead.c
```diff
@@ -0,0 +1,1374 @@
...
+}
+
+static STM32_HalStatusTypeDef stm32_aead_generate_tag(stm32_hal_aes_ctx_t *ctx,
+                                                      uint32_t *tag)
+{
+  stm32_hal_status_t hal_status;
+  STM32_HalStatusTypeDef status;
+
+  /* If no data, ensure the HAL is informed (process O bytes) */
+  if (ctx->aead_config.current_message_size == 0)
```
mathieuchopstm: Maybe:
```suggestion
  /*
   * If no data, ensure the HAL is informed (process O bytes).
   * NOTE: stm32_aead_process_buffer() locks HW mutex itself
   * so we must not acquire it yet.
   */
  if (ctx->aead_config.current_message_size == 0)
```
BTW: process `O` bytes -> process `0` bytes? or is this a well-known variable called `O`?

mathieuchopstm (commented):
A few more things. I went through the Copilot comments and closed the ones which I saw fixed. some remain open because the fix wasn't trivial enough for me to see, and some remain open because they have not been addressed yet (mainly in `rng.c`)

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aead.c
etienne-lms: It should be `0`

lib/stm32_psa_crypto_drivers/README.rst
etienne-lms: Oops!

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/src/stm32_hal_aes_private.h
etienne-lms: I messed this! Thanks for noticing.

lib/stm32_psa_crypto_drivers/stm32_psa_transparent_drivers/core/include/stm32_hal_hash.h
etienne-lms: ok

